PR #21, opened 2020-08-06
Add missing SSE instructions to X86 assembler

k-sareen:
Adds missing SSE instructions and tests for the X86 assembler. I've tried to keep to idiomatic virgil as much as possible.


aeneas/src/x86/X86Assembler.v3
```diff
@@ -1,6 +1,24 @@
...
 // See LICENSE for details of Apache 2.0 license.
 
+enum RoundingMode {
+	TO_NEAREST,
+	TO_NEG_INF,
+	TO_POS_INF,
+	TO_ZERO
+}
+
+component Rounding {
```
titzer, line 11: You can use an enum parameter for the mode so that you don't have to switch over the enum.
It works a lot like a class param.

A good example is in aeneas/src/wasm/WasmOp.v3:5:enum WasmOp(opcode: byte, text_name: string) {

k-sareen: Done. Thanks for the tip!

PR #22, opened 2020-08-10
Add ctags support

k-sareen:
I've implemented a ctags file generator for both emacs- and vi-style tags. I'm not a whizz with emacs, but in my rudimentary tests with it, the tag support was satisfactory. Please do try it out yourself to see if it works as expected.

A few quirks of the implementation:
- The generated tags file doesn't exactly work nicely with enums as I haven't made a proper `TagKind` for it. That might in a future version
- There is (at least with vi-style tags) an extra field one can add that informs it about scopes (the scope field is fairly rudimentary and doesn't really add much), but I have not implemented it yet
- Local variables have deliberately been excluded from the generated tags as the number of tags would be too large as well as the efficacy of the tags are reduced (ex: multiple local variables with the same identifier but different scopes) 
- Currently, the output file name and location are not configurable; I was trying to make it work with a `newStringOption` (basing off of `src/main/Aeneas.v3`), however it wasn't writing to the correct file name. Do you have any tips or ideas for implementing the same?
- In the process of making the tag generator, I implemented a few utility functions (namely `fold`s and `strCmp`) which I've added to the standard library. I don't use the `fold`s anymore, but I have added them to the PR regardless

Sorry for the massive PR and spaghetti code haha.

~Kunal


apps/vctags/Tags.v3
```diff
@@ -0,0 +1,119 @@
...
+		var aTag = a.formatTag();
+		var bTag = b.formatTag();
+		return Strings.strCmp(aTag, bTag) <= 0;
+	}
+
+	def formatEmacs(files: Vector<TagFile>) -> string {
+		var str = "";
+		for (i < files.length) {
+			var file = files[i], size = 0;
+			var header = Vector<byte>.new();
```
titzer, line 87: Probably better to use a StringBuffer here.

apps/vctags/vctags.v3
```diff
@@ -0,0 +1,100 @@
...
+		System.fileClose(fd);
+	}
+
+	def generateMemberTags(file: TagFile, vstCompound: VstCompound, isClass: bool) {
+		var kind: TagKind;
+		var members = Lists.toArray(vstCompound.members);
+		for (member in members) {
+			var decl = file.extractLine(member.token.beginLine);
+			if (isClass)
+				kind = if (VstMethod.?(member), TagKind.METHOD, TagKind.DEFINITION);
```
titzer, line 63: Use curlies if the if cases don't fit onto one line.

lib/util/Arrays.v3
```diff
@@ -3,53 +3,62 @@
 
 // Utility methods for arrays.
 component Arrays {
-	// get the length of {array}, but returning 0 for a null array
```
titzer, line 6: Please move lib/util changes to a separate PR.

apps/vctags/compile.sh
```diff
@@ -0,0 +1 @@
...Vector.v3 ../../aeneas/src/util/WorkQueue.v3 ../../aeneas/src/v3/Mangle.v3 ../../aeneas/src/v3/SystemModule.v3 ../../aeneas/src/v3/TypeSystem.v3 ../../aeneas/src/v3/V3Array.v3 ../../aeneas/src/v3/V3Class.v3 ../../aeneas/src/v3/V3Component.v3 ../../aeneas/src/v3/V3Enum.v3 ../../aeneas/src/v3/V3Op.v3 ../../aeneas/src/v3/V3.v3 ../../aeneas/src/vst/FpSketch.v3 ../../aeneas/src/vst/NumberParser.v3 ../../aeneas/src/vst/ParserState.v3 ../../aeneas/src/vst/Parser.v3 ../../aeneas/src/vst/Token.v3 ../../aeneas/src/vst/VarEnvironment.v3 ../../aeneas/src/vst/Verifier.v3 ../../aeneas/src/vst/VstMatcher.v3 ../../aeneas/src/vst/VstPrinter.v3 ../../aeneas/src/vst/Vst.v3 ../../aeneas/src/wasm/CfgRestructurer.v3 ../../aeneas/src/wasm/WasmCodeGen.v3 ../../aeneas/src/wasm/WasmOp.v3 ../../aeneas/src/wasm/WasmTarget.v3 ../../aeneas/src/x86/DarwinModule.v3 ../../aeneas/src/x86/LinuxModule.v3 ../../aeneas/src/x86/X86Assembler.v3 ../../aeneas/src/x86/X86Backend.v3 ../../aeneas/src/x86/X86CodeGen.v3 ../../aeneas/src/x86/X86Darwin.v3 ../../aeneas/src/x86/X86Linux.v3 ../../aeneas/src/x86/X86MacroAssembler.v3 ../../aeneas/src/x86/X86RegSet.v3 ../../aeneas/src/x86/X86Runtime.v3 ../../aeneas/src/x86/X86Target.v3
```
titzer, line 1: Instead of compile.sh, please add a DEPS file that lists the Virgil files this app depends on.
AFAICT that should be just one line:
../../aeneas/src/*/*.v3

lib/util/Arrays.v3
k-sareen: Do you want the `src/util/Arrays.v3` and `src/util/Strings.v3` changes in that PR as well?

apps/vctags/Tags.v3
k-sareen: Specifically for `str` or for the variables of type `Vector<byte>` as well?

apps/vctags/compile.sh
k-sareen: Done

lib/util/Arrays.v3
k-sareen: I've currently just reverted `lib/util/Arrays.v3` to its original state. If you want the above two in a different PR as well, then I think it'll be easier (for both of us) to close this PR and make new ones.

apps/vctags/Tags.v3
k-sareen: I've currently changed everything to be a `StringBuffer`.

k-sareen:
There was a merge conflict due to the addition of a space in my branch. I've resolved it by picking my version.

PR #23, opened 2020-08-26
Add scripts to check differences in bootstrapped binary

k-sareen:
Adds two scripts:

1. `test/check-bootstrap.bash`: runs two different versions of Aeneas and outputs the ten smallest binaries which have differences (for ease of debugging).
2. `test/compare.bash`: runs two different versions of Aeneas on a particular test case and outputs the diff of the SSA, instructions, and binary. To be used in conjunction with the above.

~Kunal

### DIFF
--- a/test/check-bootstrap.bash
+++ b/test/check-bootstrap.bash
@@ -1,25 +1,19 @@
 #!/bin/bash
 
-function line() {
-    echo ================================================================================
-}
-
 SOURCE="${BASH_SOURCE[0]}"
 while [ -h "$SOURCE" ]; do
-  DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
-  SOURCE="$(readlink "$SOURCE")"
-  [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE"
+    DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
+    SOURCE="$(readlink "$SOURCE")"
+    [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE"
 done
 DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
 
-if [ $# != 3 ]; then
+. $DIR/common.bash check-bootstrap
+
+if [ $# -lt 3 ]; then
     echo "Usage: check-bootstrap.bash <target> <Aeneas before> <Aeneas after>"
     exit 1
 fi
-
-BLUE='\033[0;34m'
-GREEN='\033[0;32m'
-NC='\033[0m'
 
 target=$1
 shift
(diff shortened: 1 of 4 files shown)


test/check-bootstrap.bash
```diff
@@ -0,0 +1,64 @@
...
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do
+  DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
+  SOURCE="$(readlink "$SOURCE")"
+  [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE"
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" >/dev/null 2>&1 && pwd )"
+
+if [ $# != 3 ]; then
+    echo "Usage: check-bootstrap.bash <target> <Aeneas before> <Aeneas after>"
```
titzer, line 16: what about accepting a glob of filenames (e.g. execute/\*.v3) as the rest of the arguments, and defaulting to ./\*.v3, like the test.bash scripts do?

test/compare.bash
```diff
@@ -0,0 +1,46 @@
...
+    exit 1
+fi
+
+BLUE='\033[0;34m'
+NC='\033[0m'
+
+target=$1
+shift
+
+TEST=$1
```
titzer, line 18: nit: I wouldn't normally shift off arguments unless I was planning to use $@ later.

test/check-bootstrap.bash
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+function line() {
```
titzer, line 3: What about:

- pull line() into common.bash
- doing what diagnose.bash does
- pulling execute() out of diagnose.bash into common.bash
- put BLUE there

?
- use

titzer (changes requested):

test/check-bootstrap.bash
k-sareen: I'm not sure what you mean by "use" in the last point

test/check-bootstrap.bash
k-sareen: Done.

test/check-bootstrap.bash
k-sareen: What're your thoughts on the latest version?

titzer: Sorry, I didn't finish the comment. I meant to write "use the execute function for echoing a command before running it.

Other than that, this looks good.

k-sareen: If we use the `execute` function, stdout will get polluted with the names of all the tests and the test results (for example if we ran `execute/*.v3). Do you want to have that? The current version pipes it all out to `/dev/null` to reduce the clutter.

titzer: Ok, I guess it's alright.

PR #26, opened 2020-09-01
Refactor SSE instructions

k-sareen:
Refactors the comparison instructions to reduce repeated code. Code is more in-line with `cmov()` now. Also updates the move instructions to reflect the actual operands they can take.


k-sareen:
Note that there does not seem to be an analogue for `movd xmm, r/m32`, i.e there is no `movd r/m32, xmm`.

k-sareen:
Further note: there is a some dead code in `test/x86/asm/X86AssemblerTestGen.v3` now; namely: `do_rm_s()`, and `do_m_s()`. We can remove it if you prefer.

titzer (approved):
lgtm, nice

PR #28, opened 2020-09-12
Improve support for SSE instruction generation

k-sareen:
This is a fairly big patch. I've tried to make each commit small-ish so that reviewing it is more digestible. The meat of the patch is the implementation of conversion and truncation between floats and (unsigned) ints. 

A few things to note:
- The conversion and truncation functions added in the MacroAssembler were either implemented in C first and compiled with `icc` using `-msse4.2 -m32 -O3` or they were based on how the JVM implements them.
- The two x87 FPU instructions were what `icc` spat out for the conversion between a `float` and `unsigned int`. The conversion is quite simple to do with these instructions. It's a bit of a nightmare in just SSE instructions (although technically, the `fisttp` instruction was added in SSE3).
- I was not so sure how to respect calling convention when dealing with stub functions, so I opted to use the stack for passing values around. I was a bit stingy some places and directly return value in the SSE scratch register (`xmm7`). Please do tell me if I should change either of the above.
- Originally I had opted to have function names similar to Intel instructions, such as `cvtsui2ss` instead of `cast_ui2f` for converting a single unsigned integer to a single-precision float (sort of similar to what `v8` does). I have opted to have the simpler naming scheme over the alphabet soup. Do you have any preference?

I have completed support for single-precision floating point, passing all relevant tests. I'll send the patch for that as soon as this supporting patch is merged.

~Kunal

### DIFF
--- a/aeneas/src/x86/X86MacroAssembler.v3
+++ b/aeneas/src/x86/X86MacroAssembler.v3
@@ -78,26 +78,22 @@ class X86MacroAssembler extends X86Assembler {
 		if (s == d) return;
 		if (SSEReg.?(d)) return movsd_s_sm(SSEReg.!(d), s);
 		if (SSEReg.?(s)) return movsd_sm_s(d, SSEReg.!(s));
-		if (scratch != null) {
-			movsd_s_sm(scratch, s);
-			movsd_sm_s(d, scratch);
-		}
-		// push(s);            // push value from source memory onto the stack
-		// pop(SSEAddr.!(d));  // pop value off stack into destination memory
+		movsd_s_sm(scratch, s);
+		movsd_sm_s(d, scratch);
 	}
 	// a macro to move a value into a location
 	def movd_l_val(frame: MachFrame, loc: int, val: Val) {
-		if (loc >= X86MachRegs.XMM0 && loc <= X86MachRegs.XMM7
-				|| Float32Val.?(val)) {
+		if (loc >= X86MachRegs.XMM0 && loc <= X86MachRegs.XMM7) {
 			movd_l_fval(frame, loc, val);
 		} else {
 			var d = loc_rm(frame, loc);
 			match (val) {
 				x: Addr => {
-						 var pos = buffer.pos;
-						 movd_rm_i(d, X86Addrs.ABS_CONST);
-						 recordPatch(pos, x);
+					var pos = buffer.pos;
+					movd_rm_i(d, X86Addrs.ABS_CONST);
+					recordPatch(pos, x);
 				}
+				x: Float32Val => movd_l_fval(frame, loc, x); // TODO: Float64Val case
 				_ => movd_rm_i(d, V3.unboxIntegral(val));
 			}
 		}
@@ -109,7 +105,7 @@ class X86MacroAssembler extends X86Assembler {
 		if (val != null) {
 			match (val) {
 				x: Float32Val => f = int.view(x.bits);
-				x: Float64Val => f = int.view(x.bits);
+				x: Float64Val => f = int.view(x.bits); // TODO: fix Float64Val case
 			}
 		}
 		movd_rm_i(X86Regs.EBP, f);
@@ -140,11 +136,7 @@ class X86MacroAssembler extends X86Assembler {
 			X86MachRegs.EDI => return edi;
 			X86MachRegs.EBP => return ebp;
 		}
-		var regSet = frame.conv.regSet, wordSize = mach.data.addressSize, offset: int;
-		if (loc >= regSet.calleeStart) offset = wordSize * (loc - regSet.calleeStart);
-		else if (loc >= regSet.callerStart) offset = frame.size() + (wordSize * (loc - regSet.callerStart));
-		else if (loc >= regSet.spillStart) offset = wordSize * (loc - regSet.spillStart + frame.spillArgs);
-		else return failLocation("invalid spill location", loc, frame.conv.regSet);
+		var offset = calc_offset(frame, loc, false);
 		return esp.plus(offset);
 	}
 	def failLocation(msg: string, loc: int, regSet: MachRegSet) -> X86Reg {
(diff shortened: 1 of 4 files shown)


k-sareen:
Don't merge the patch yet, sorry. I would also like to add support for the FloatBitEq operator to this patch.

aeneas/src/x86/X86MacroAssembler.v3
```diff
@@ -51,22 +73,51 @@ class X86MacroAssembler extends X86Assembler {
...
+	// a macro to move between any two SSE register/memory operands
+	def movsd_sm_sm(d: SSERm, s: SSERm, scratch: SSEReg) {
+		if (s == d) return;
+		if (SSEReg.?(d)) return movsd_s_sm(SSEReg.!(d), s);
+		if (SSEReg.?(s)) return movsd_sm_s(d, SSEReg.!(s));
+		if (scratch != null) {
+			movsd_s_sm(scratch, s);
+			movsd_sm_s(d, scratch);
+		}
+		// push(s);            // push value from source memory onto the stack
```
titzer, line 85: Do we still need the case where no scratch register is available?

aeneas/src/x86/X86MacroAssembler.v3
```diff
@@ -51,22 +73,51 @@ class X86MacroAssembler extends X86Assembler {
...
-		var d = loc_rm(frame, loc);
-		if (val == null) {
-			movd_rm_i(d, 0);
-		} else if (Addr.?(val)) {
-			var pos = buffer.pos;
-			movd_rm_i(d, X86Addrs.ABS_CONST);
-			recordPatch(pos, Addr.!(val));
+		if (loc >= X86MachRegs.XMM0 && loc <= X86MachRegs.XMM7
+				|| Float32Val.?(val)) {
+			movd_l_fval(frame, loc, val);
```
titzer, line 92: What about moving the Float32Val case down into the match below? You'll need a Float64Val case too. It can be the case that the constant is null (indicating the default value for the type, which should be handled in the match too.

aeneas/src/x86/X86MacroAssembler.v3
```diff
@@ -51,22 +73,51 @@ class X86MacroAssembler extends X86Assembler {
...
 		}
 	}
+	// a macro to move a value into an SSE location
+	def movd_l_fval(frame: MachFrame, loc: int, val: Val) {
+		var f = 0;
+		var d = loc_sm(frame, loc);
+		if (val != null) {
+			match (val) {
+				x: Float32Val => f = int.view(x.bits);
+				x: Float64Val => f = int.view(x.bits);
```
titzer, line 112: `int.view` will slice off only the low 32-bits. You'll have to handle doubles differently here.

aeneas/src/x86/X86MacroAssembler.v3
```diff
@@ -100,6 +151,44 @@ class X86MacroAssembler extends X86Assembler {
...
+			X86MachRegs.XMM1 => return xmm1;
+			X86MachRegs.XMM2 => return xmm2;
+			X86MachRegs.XMM3 => return xmm3;
+			X86MachRegs.XMM4 => return xmm4;
+			X86MachRegs.XMM5 => return xmm5;
+			X86MachRegs.XMM6 => return xmm6;
+			X86MachRegs.XMM7 => return xmm7;
+		}
+		var regSet = frame.conv.regSet, wordSize = mach.data.addressSize, offset: int;
+		if (loc >= regSet.calleeStart) offset = wordSize * (loc - regSet.calleeStart);
```
titzer, line 182: Is it possible to factor out the offset calculation, since it doesn't depend on the return value being an SSERm or X86Rm?

aeneas/src/x86/X86MacroAssembler.v3
```diff
@@ -320,6 +409,204 @@ class X86MacroAssembler extends X86Assembler {
...
+			fld_d(save);
+		}
+		fisttp_q(save);
+		movd_r_rm(dest, save.toX86Addr());
+		add.rm_i(esp, 0x8);
+	}
+	def cast_d2ui(dest: X86Reg, a: SSERm) {
+		// TODO: convert double to unsigned int
+	}
+	def f2ui_fixup() {
```
titzer, line 430: I don't quite understand this routine. For float/double convert to u32, you should be able to do a convert to i64 and then compare that result (unsigned 64-bit >=) with u32.max, no?

k-sareen: That's essentially what it's doing. Basically the function is:
```
def f2ui(float f) -> u32 {
  if( f <= 0) return 0;
  if (f >= 4294967296) return 0xffffffff
  return u32.!(f);
}
```
Arguably, the first if condition is not required and we could do with removing it.

aeneas/src/x86/X86MacroAssembler.v3
k-sareen: I have currently not implemented support for doubles. We can remove it for the time being, if you prefer.

aeneas/src/x86/X86MacroAssembler.v3
k-sareen: I don't imagine so. X86CodeGen currently explicitly supplies `xmm7` as the scratch register. I'll remove the comments.

aeneas/src/x86/X86MacroAssembler.v3
k-sareen: Maybe? The issue is the `else return failLocation...`. How do we nicely implement that. I'm thinking we could internally output `(int, bool)` for calculating the offset with the `bool` representing if the location has been found and the offset calculated?

k-sareen: Done.

aeneas/src/x86/X86MacroAssembler.v3
k-sareen: I've added a TODO as a reminder that it needs to be fixed.

aeneas/src/x86/X86MacroAssembler.v3
```diff
@@ -320,6 +411,204 @@ class X86MacroAssembler extends X86Assembler {
 		}
 		if (frame != null) mach.runtime.src.recordFrameEnd(codeOffset());
 	}
+
+	def cast_f2ui(dest: X86Reg, a: SSERm) {
```
titzer, line 415: Let's avoid the name "cast" here to prevent confusion with source- and IR- level cast operations, which can throw exceptions.

aeneas/src/x86/X86MacroAssembler.v3
titzer: I meant, compare the result of the conversion using an unsigned 64-bit integer comparison.

titzer: Also, this function needs some documentation, ~2 sentences about what the inputs/outputs are and what it does.

titzer (approved):

k-sareen:
Ah. I was refactoring some of the code to be similar to what the JVM does. I suppose I'll make a separate patch removing all the x87 dependencies when I'm done with the refactor.

PR #29, opened 2020-09-14
Add single-precision float support

k-sareen:
Adds single-precision floating point support. Hell yeah!

Most of the stuff for doubles should already be part of this as well. Hopefully that doesn't take as long to complete. I'll get around to removing the x87 dependencies in a future patch.

~Kunal

### DIFF
--- a/aeneas/src/x86/X86CodeGen.v3
+++ b/aeneas/src/x86/X86CodeGen.v3
@@ -14,7 +14,7 @@ component X86Addrs {
 	def REL_CONST = 0x44332205;
 
 	def ABS_PATCH = X86Addr.new(null, null, 1, 0x44332211);
-	def ABS_PATCH_SSE = SSEAddr.new(null, null, 1, 0x44332211);
+	def ABS_PATCH_SSE = ABS_PATCH.toSSEAddr();
 }
 // Pattern match of a comparison between two vars or a var and a constant
 class X86CmpMatch {
@@ -224,12 +224,12 @@ class X86CodeGen extends MachCodeGen {
 				if (isDouble) gen_sse_op2("divsd", X86Assembler.divsd, v, i.input0(), i.input1());
 				else gen_sse_op2("divss", X86Assembler.divss, v, i.input0(), i.input1());
 			}
-			FloatBitEq(isDouble) => gen_sse_bit_eq(v, i.input0(), i.input1(), isDouble);
-			FloatEq(isDouble) => gen_sse_cmp(sseCond(i.op.opcode), v, i.input0(), i.input1(), isDouble);
-			FloatNe(isDouble) => gen_sse_cmp(sseCond(i.op.opcode), v, i.input0(), i.input1(), isDouble);
-			FloatLt(isDouble) => gen_sse_cmp(sseCond(i.op.opcode), v, i.input0(), i.input1(), isDouble);
-			FloatLteq(isDouble) => gen_sse_cmp(sseCond(i.op.opcode), v, i.input0(), i.input1(), isDouble);
-			FloatAbs(isDouble) => gen_sse_fabs(v, i.input0(), isDouble);
+			FloatBitEq(isDouble) =>	gen_sse_bit_eq(v, i.input0(), i.input1(), isDouble);
+			FloatEq(isDouble) => 	gen_sse_cmp(X86Conds.Z, v, i.input0(), i.input1(), isDouble);
+			FloatNe(isDouble) => 	gen_sse_cmp(X86Conds.NZ, v, i.input0(), i.input1(), isDouble);
+			FloatLt(isDouble) => 	gen_sse_cmp(X86Conds.C, v, i.input0(), i.input1(), isDouble);
+			FloatLteq(isDouble) => 	gen_sse_cmp(X86Conds.NA, v, i.input0(), i.input1(), isDouble);
+			FloatAbs(isDouble) => 	gen_sse_fabs(v, i.input0(), isDouble);
 			FloatCeil(isDouble) => {
 				if (isDouble) gen_sse_round("roundsd_ceil", X86Assembler.roundsd, v, i.input0(), RoundingMode.TO_POS_INF);
 				else gen_sse_round("roundss_ceil", X86Assembler.roundss, v, i.input0(), RoundingMode.TO_POS_INF);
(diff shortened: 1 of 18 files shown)


aeneas/src/x86/X86CodeGen.v3
```diff
@@ -596,12 +686,28 @@ class X86CodeGen extends MachCodeGen {
 		}
 		return (i, 1);
 	}
+	def isSigned(i: SsaApplyOp, isReturn: bool) -> bool {
+		if (isReturn) return IntType.!(i.getType().typeCon.singleton).signed;
```
titzer, line 690: You shouldn't need to go to .typeCon.singleton; that just gets you back to the same type.

aeneas/src/x86/X86CodeGen.v3
```diff
@@ -636,13 +743,30 @@ class X86CodeGen extends MachCodeGen {
...
 		var lt = matchLoad(t, mach.sizeOf(t));
 		gen(lt.0, asm_loada, (lt.1, dfngpr(dest), addr));
 	}
 	def gen_loada_disp(dest: VReg, t: Type, addr: Addr, disp: SsaInstr) {
+		if (t.typeCon.kind == V3Kind.FLOAT) return gen_loada_disp_sse(dest, t, addr, disp);
 		var lt = matchLoad(t, mach.sizeOf(t));
 		gen(lt.0, asm_loada_disp, (lt.1, dfngpr(dest), addr, gpr(disp)));
 	}
+	def gen_loadx_sse(dest: VReg, t: Type, base: SsaInstr, scale: byte, disp: int, canTrap: bool, source: Source) {
+		if (mach.sizeOf(t) == 4) {
```
titzer, line 756: Can you factor out {mach.sizeOf(t) == 4} to an {isDouble()} routine and invert these cases?

aeneas/src/x86/X86CodeGen.v3
```diff
@@ -670,6 +794,21 @@ class X86CodeGen extends MachCodeGen {
 		m(asm, r(dest), r(disp).plus(X86Addrs.ABS_CONST));
 		recordPatch(pos, addr);
 	}
+	def asm_loadx_sse(m: (X86Assembler, SSEReg, SSERm) -> void, dest: int, base: int, scale: byte, disp: int, canTrap: bool, source: Source) {
+		var off = asm.codeOffset();
+		m(asm, s(dest), SSEAddr.new(null, r(base), scale, disp));
+		if (canTrap && rtsrc != null) rtsrc.recordSource(off, source);
```
titzer, line 800: That reminds me...we should have stack trace tests for OOB/nullchecks involving floats and doubles. Can you drop a TODO somewhere (I guess here is fine) that we need to add that to the stacktrace tests? It's a bit tricky because floating point is still an optional feature and the stacktrace tests don't deal with that yet.

aeneas/src/x86/X86CodeGen.v3
```diff
@@ -379,8 +453,10 @@ class X86CodeGen extends MachCodeGen {
 		var ptr = i.inputs[0], ssaval = i.input1();
 		if (SsaConst.?(ptr.dest)) {
 			// store to absolute address
+			var t = i.op.typeArgs[1];
 			var addr = Addr.!(val(ptr.dest));
-			gen_storex_abs(mach.sizeOf(i.op.typeArgs[1]), addr, ssaval);
+			if (t.typeCon.kind == V3Kind.FLOAT) gen_storex_abs_sse(mach.sizeOf(t), addr, ssaval);
```
titzer, line 458: Similarly, this should probably use {isDouble()} and make it impossible to have an incorrect size. It might be better to do the if (isDouble) check here instead of lower down, too.

aeneas/src/x86/X86CodeGen.v3
```diff
@@ -168,6 +169,79 @@ class X86CodeGen extends MachCodeGen {
...
+			FloatMul(isDouble) => {
+				if (isDouble) gen_sse_op2("mulsd", X86Assembler.mulsd, v, i.input0(), i.input1());
+				else gen_sse_op2("mulss", X86Assembler.mulss, v, i.input0(), i.input1());
+			}
+			FloatDiv(isDouble) => {
+				if (isDouble) gen_sse_op2("divsd", X86Assembler.divsd, v, i.input0(), i.input1());
+				else gen_sse_op2("divss", X86Assembler.divss, v, i.input0(), i.input1());
+			}
+			FloatBitEq(isDouble) => gen_sse_bit_eq(v, i.input0(), i.input1(), isDouble);
+			FloatEq(isDouble) => gen_sse_cmp(sseCond(i.op.opcode), v, i.input0(), i.input1(), isDouble);
```
titzer, line 228: I think you know the condition from the operator here, so no need to call sseCond, right?

test/float/test.bash
```diff
@@ -2,7 +2,7 @@
 
 V3C_OPTS="$V3C_OPTS -fp"
 RUN_INT=0
-RUN_NATIVE=0
+RUN_NATIVE=1
```
titzer, line 5: Probably don't want to flip that until all the tests, including doubles, pass.

aeneas/src/x86/X86CodeGen.v3
```diff
@@ -740,6 +899,87 @@ class X86CodeGen extends MachCodeGen {
 		var f = if(commutative, asm_op2_comm, asm_op2);
 		gen(name, f, (m, dfngpr(dest), use(a), use(b), hint(a, dest)));
 	}
+	// ---------- Basic operations (SSE) ----------------------------------------
+	def gen_sse_op2(name: string, m: (X86Assembler, SSEReg, SSERm) -> void, dest: VReg, a: SsaInstr, b: SsaInstr) {
+		gen(name, asm_sse_op2, (m, dfnsse(dest), use(a), use(b), hint(a, dest)));
+	}
+	def asm_sse_op2(m: (X86Assembler, SSEReg, SSERm) -> void, dest: int, a: int, b: int, v: void) {
+		dest = loc(dest); a = loc(a); b = loc(b);
+		if (dest != a) {
```
titzer, line 908: single-line if for consistency with surrounding code

aeneas/src/x86/X86CodeGen.v3
k-sareen: Yep. This was before the `isDouble` patch, so I had them fall through before. I'll change that now.

aeneas/src/x86/X86CodeGen.v3
k-sareen: Ah. My bad. Fixed.

k-sareen:
Ah damn. Seems like I really fudged the git rebase. Let's just close this in favour of making a new one? Would you again prefer to have two different patches for single and double-precision floats?

titzer:
Can you factor out all changes that are not the codegen into one PR and then all the codegen changes into another?

k-sareen:
Sure.

PR #30, opened 2020-09-24
Improve floating point support for x86 target

k-sareen:
This patch basically condenses every change *except* the actual isel for floating point support on x86. Unfortunately, I looked into it and the x87 FPU dependencies will be necessary until we move to x64.

I've heavily commented the stubs and macros in the X86MacroAssembler to make it clear what every function does. I'll create a new PR for the isel when this one is approved.

~Kunal


k-sareen:
Haha damn; almost always forget to bump the Aeneas version!

titzer (approved):
Whew! Nice job!

PR #31, opened 2020-09-24
Add floating point support for x86 target

k-sareen:
This patch finally adds floating point support for x86 target to the compiler! Hell yeah! I'll make more test cases particularly stressing the register allocator and spills.

Wew! :D

~Kunal


titzer:
Great work!

Wizard can successfully pass Wasm spec tests with this patch. However, Aeneas crashes when trying to compile Wizard's unit tests.

!TypeCheckException
	in X86MacroAssembler.movq_m_m() [/home/titzer/virgil/aeneas/src/x86/X86MacroAssembler.v3 @ 110:49]
	in X86CodeGen.asm_movd_l_l() [/home/titzer/virgil/aeneas/src/x86/X86CodeGen.v3 @ 1464:45]
	in MoveResolver.orderMove() [/home/titzer/virgil/aeneas/src/mach/MoveResolver.v3 @ 113:43]

k-sareen:
Hmm. Somehow we're getting a value there which is not an X86Addr. Are you perhaps trying to move a value in memory to a value (such as an `i64) which is in two registers?

Could you point me to the unit test?

titzer (approved):
Nice, I think we can land this now.

PR #33, opened 2020-09-30
Use fpScratch as a buffer for SSE instrs with two inputs

k-sareen:


titzer (commented):
Sorry, I should have caught this before in review. This looks correct.

Please add test cases that demonstrates the bug being fixed here.

titzer (approved):
lgtm, thanks

PR #36, opened 2020-11-24
Implement test progress printer in C

k-sareen:
Finished my implementation of the `progress` pretty printer in C. It should be a drop-in replacement for the one in Virgil. The test suite just compares the output generated with the expected output  (generated from the original Virgil `progress` executable on x86-linux). Runs without generating any asan and ubsan errors at least for the test set provided.

I haven't changed the test suite to use this version of `progress` as from memory you had a local branch that did most of the work. I also haven't deleted the Virgil version, as I thought that would be up to your discretion.

P.S. the manual memory management hacks were ugly :/


apps/progress/progress.c
```diff
@@ -0,0 +1,359 @@
...
+                    break;
+            }
+        }
+    }
+
+    if (mode == SUMMARY) indent = 0;
+
+    report_start();
+    while (1) {
+        char v[1];
```
titzer, line 68: Character-by-character is fine for now; I just did it that way in Virgil because I was too lazy to make multi-character reads from System.* reliably detect the number of available characters available and read only those (soooo rusty with Unix pipes). Can you add a TODO to do a multi-character read here?

titzer, line 68: You could read directly into the right position in the line_buffer here.

apps/progress/progress.c
```diff
@@ -0,0 +1,359 @@
...
+
+void outc(char c)
+{
+    printf("%c", c);
+    char_backup++;
+}
+
+void outi(int i)
+{
+    printf("%d", i);
```
titzer, line 123: printf returns the number of characters printed, so you don't have to calculate it manually.

apps/progress/progress.c
```diff
@@ -0,0 +1,359 @@
...
+    char_backup++;
+    while (i >= 10) {
+        char_backup++;
+        i /= 10;
+    }
+}
+
+void outs(char *str)
+{
+    printf("%s", str);
```
titzer, line 133: Same as above.

apps/progress/progress.c
```diff
@@ -0,0 +1,359 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include "progress.h"
+
+void insert(struct failure *f)
+{
```
titzer, line 8: Can you put opening curlies on the previous line?

titzer (commented):
Great, thanks for doing that!

It's fine to add this code in apps/ for now, but I want to start factoring common code out into an actual standard library, and this is a good candidate for that. Alternatively, we could move it into test/bin/src/ with other testing-related code (along with the Virgil version).

apps/progress/progress.c
k-sareen: Not a fan of K&R's style eh? ;)

apps/progress/progress.c
k-sareen: Just implemented multicharacter using `getline`.

k-sareen: See above

k-sareen:
Where would you ideally put it, i.e. what is 'standard library' for Virgil?

titzer (approved):
lgtm, thanks

PR #37, opened 2020-12-08
Fix comparison against string constant

k-sareen:
The `$1` should be quoted, else the comparison doesn't go through.


titzer (approved):
lgtm

PR #41, opened 2021-02-28
Add check for word size when calculating number of spill slots

k-sareen:
Quick change that I noticed. Makes it so that 64-bit only uses one spill slot. Does make assumptions that at least 32-bit word size.

### DIFF
--- a/aeneas/src/mach/MachBackend.v3
+++ b/aeneas/src/mach/MachBackend.v3
@@ -91,8 +91,7 @@ class MachFrame(conv: MachCallConv, wordSize: int) {
 	def allocSpill(regClass: RegClass) -> int {
 		match (regClass) {
 			I64, F64 => {
-				var numSpill = 64 / wordSize;
-				if (numSpill == 0) numSpill = 1; // in case word size is larger than 64-bit XXX: is it required?
+				var numSpill = if(wordSize >= 64, 1, 2); // XXX: assumes at least 32-bit word size
 				var spill = conv.regSet.spillStart + spillVars;
 				spillVars += numSpill;
 				return spill | IS_64;
--- a/aeneas/src/mach/MachBackend.v3
+++ b/aeneas/src/mach/MachBackend.v3
@@ -81,7 +81,7 @@ class VReg {
 	}
 }
 // Representation of a machine frame, including spill slots and space for outgoing params
-class MachFrame(conv: MachCallConv) {
+class MachFrame(conv: MachCallConv, wordSize: int) {
 	var spillVars: int;		// spilled variables
 	var spillArgs: int;		// space for outgoing spilled arguments
 	var frameSize: int = -1;	// architecture-specific total frame size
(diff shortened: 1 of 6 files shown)


aeneas/src/x86-64/X86_64Backend.v3
```diff
@@ -220,7 +220,7 @@ class X86_64Backend extends MachBackend {
 		asm.xaddq_m_r(addr, reg);
 	}
 	def getFrame(ssa: SsaGraph) -> MachFrame {
-		return MachFrame.new(X86_64VirgilCallConv.getForGraph(ssa));
+		return MachFrame.new(X86_64VirgilCallConv.getForGraph(ssa), mach.code.addressSize);
```
titzer, line 223: Should probably use the data.addressSize here as in other places.

aeneas/src/x86/X86CodeGen.v3
```diff
@@ -55,7 +55,7 @@ class X86CodeGen extends OldCodeGen {
 	var kernel: Kernel;
 
 	new(mach: MachProgram, context: SsaContext) super(mach, context) {
-		frame = MachFrame.new(X86VirgilCallConv.getForGraph(context.graph));
+		frame = MachFrame.new(X86VirgilCallConv.getForGraph(context.graph), mach.code.addressSize);
```
titzer, line 58: Here, too.

aeneas/src/mach/MachBackend.v3
```diff
@@ -91,8 +91,7 @@ class MachFrame(conv: MachCallConv, wordSize: int) {
 	def allocSpill(regClass: RegClass) -> int {
 		match (regClass) {
 			I64, F64 => {
-				var numSpill = 64 / wordSize;
-				if (numSpill == 0) numSpill = 1; // in case word size is larger than 64-bit XXX: is it required?
+				var numSpill = if(wordSize >= 64, 1, 2); // XXX: assumes at least 32-bit word size
```
titzer, line 94: An alternative would be to pass an {Alignment} object. There's one in {AddressSpace.addrAlign}; then you can do {align.alignUp({4, 8}) >> align.shift).

k-sareen: You mean pass it to `allocSpill()` or during creation as a parameter to `MachFrame`?

titzer: Upon creation.

k-sareen: Done. But slight aside is that I think we're assuming that `RegClass.REF` is always 32bits/4bytes as opposed to being architecture dependent.

titzer: Yeah, I was going to suggest adding refSize as an argument too. It's in the MachProgram.

k-sareen: I wonder if it's more elegant to pass the `AddressSpace` around instead. Everything we want is encapsulated by it (`addressSize` and `addrAlign`).

k-sareen: Also do we care about annotating everything with `IS_64` on a 64-bit architecture? It doesn't really matter as the spill size is 64-bits so we don't have the special case anymore.

titzer: The refSize is (or will be) independent of the address space size, as I plan to implement compressed references on 64-bit to save memory.

I think we can get rid of IS_64 in the future when we deprecate the OldCodeGen. In the new code gen, there is always a VReg available when generating moves, which has a register class. I think the main culprit for that mess was the MoveResolver, which is kind of folded into SimpleRegAlloc now.

k-sareen: Cool. I've just pushed the small fix then. I think the PR should be good to go.

PR #44, opened 2021-08-17
Escape special characters for vi-style tags

k-sareen:
I've been sitting on this one for more than a year I think -- I just realized you've started using the ctags more often so decided to make a PR for this. Super hacky way to escape characters though


apps/vctags/Tags.v3
```diff
@@ -50,6 +53,17 @@ class Tag {
 
 		return id.toString();
 	}
+	// functions to escape '/' and '$' characters
+	private def escapeChars(s: string) -> string {
+		return Arrays.foldR(Arrays.map(s, escapeChar), Strings.concat, "");
```
titzer, line 58: That's pretty inefficient. What about using a StringBuilder and a for loop?

titzer (commented):
Nice. Looks good other than a small efficiency increase.

apps/vctags/Tags.v3
k-sareen: Done

PR #56, opened 2022-06-06
[x86-64-darwin] Fix fatal stub handlers

refusal-type-bamboo:
Fixed stub handlers and binary format.


titzer (approved):
lgtm, after testing locally.

PR #70, opened 2022-07-14
Started DSL project

refusal-type-bamboo:
Started implementing the parser for layouts.
Layout initialization not added.


titzer (approved):
Looks good, can you bump the version number with 'aeneas bump'?

refusal-type-bamboo:
Updated. Sorry for that!

aeneas/src/vst/Parser.v3
```diff
@@ -132,6 +132,14 @@ component Parser {
 					return true;
 				}
 			}
+			'l' => {
```
titzer, line 135: I had a look back at our whiteboard conversation, and I think we should use the 'struct' keyword for fixed layouts. Can you update this PR to use 'struct' for the keyword and name the new Vst node `VstStruct`?

titzer (changes requested):

aeneas/src/vst/Parser.v3
refusal-type-bamboo: Got it. All keyword 'layout' changed to 'struct', methods renamed.

titzer (approved):
lgtm, thanks!

PR #76, opened 2022-08-04
Parser modified, test cases added, verifier unchanged

refusal-type-bamboo:
Hi Dr.Titzer,

Below are the changes to the parser mentioned before. Changes to the verifiers are not included and I've tested and compiled the code and it runs. Though I'm currently unable to start working on the changes to the verifier, let me know if there's any issue with the parser and I can certainly fix them as soon as possible.

Thanks!
Yunqing


titzer (commented):
Thanks, will merge now and do some cleanups.

PR #84, opened 2022-08-31
Parsing added

refusal-type-bamboo:
Hi Dr.Titzer,

Below is a summary of work done:
- modified the test cases according to the new syntax (`inline` deleted)
- parsing `rephint` on top level of struct
- parsing `rephint` for struct fields
- structure for checking OOB error and field overlapping

For `Refs` and exact details of checking field decl error, I found my previous approach incorrect and feel like it would be better if `resolvetype`(especially the 'searching for struct first' part) is implemented before start checking field overlapping. 

Let me know if there's any issue.
Yunqing


titzer:
I've now merged this in as a separate patch.

PR #90, opened 2022-09-09
[doc] errata

srackham:


titzer (approved):
lgtm. I should probably run my examples through the compiler before adding them to the doc :-)

PR #117, opened 2022-12-23
[x86_64-linux]: Add basic networking support

k-sareen:
This commit adds basic networking support to the x86_64 Linux backend. This commit only implements client sockets.


k-sareen:
Pretty much a WIP. Wanted to get your thoughts before I added more support. Here's a sample program to test this:

```
def main() {
        def localhost = System.LOCALHOST_V4;
        def sock = ClientSocket.new(localhost, 5555, 80);
        def buffer = Array<byte>.new(512);
        System.puts("Hello ");
        System.puti(sock.fd);
        System.ln();
        sock.connect();
        sock.write("Testing\n");
        System.puti(sock.read(buffer));
        System.ln();
        System.puts(buffer);
        sock.close();
}
```

In a separate terminal run `ncat -e /bin/cat -k -l 5555 -v` and you should see `Testing` print in the terminal running Virgil code.

k-sareen:
Couple of things:

 1. I'm not happy with how `InetAddress` is implemented as there is no size for the arrays. Since `InetAddress` is a public type, theoretically anyone can create a faulty IP address. Ideally it's a `u8[4]` and `u8[16]` array for IPv4 and v6 respectively. I have a proposed solution in the comments, but I've just opted for this as it is currently easier to work with.
 2. I'll add parsing support for IPv4 and IPv6 addresses later. It might be better to directly use a `host: string` field for the constructor in `SocketBase` instead of the naked `InetAddress`. This depends on the API we want to provide to the user.
 3. Currently there is a fairly disgusting hack to marshal the data across the kernel. Is there a better way to do this other than `struct`s which aren't stable yet, from my understanding.

rt/x86-64-linux/System.v3
```diff
@@ -8,7 +8,120 @@
...
+	return u16.view(((n & 0xFF) << 8) | ((n & 0xFF00) >> 8));
+}
+
+def u32ToNetworkByteOrder(n: u32) -> u32 {
+	return u32.view(((n & 0xFF) << 24) | ((n & 0xFF00) << 8) |
+		((n & 0xFF0000) >> 8) | ((n & 0xFF000000) >> 24));
+}
+
 component System {
+	def ANY_V4 = InetAddress.Inet4([0, 0, 0, 0]);
```
titzer, line 119: I think it'd be best not to add anything to `System` just yet. So far it's just the minimal portable stuff to get off the ground. It feels like this should probably live under `lib/net` for now. It's platform-specific, so maybe `lib/net/x86-64-linux` until we can figure out a nicer way to make portable interfaces that automatically build against the right target library.

rt/x86-64-linux/System.v3
```diff
@@ -8,7 +8,120 @@
...
+				//   struct in_addr     sin_addr;    // Internet address
+				//   unsigned char      sin_zero[8]; // Same size as struct sockaddr
+				// };
+				def sockAddr = Array<byte>.new(LinuxConst.SOCKADDR_V4_SIZE);
+				def nboPort = u16ToNetworkByteOrder(u16.view(port));
+				def nboAddr = u32ToNetworkByteOrder(
+						u32.view((int.view(addrV4[0]) << 24) | (int.view(addrV4[1]) << 16)
+							| (int.view(addrV4[2]) << 8) | addrV4[3])
+				);
+				// XXX: Disgusting hack alert! Hopefully this can be cleaned up with structs
```
titzer, line 73: Another way to do this is with a `DataWriter`.

rt/x86-64-linux/System.v3
```diff
@@ -8,7 +8,120 @@
...
+type InetAddress {
+	case Inet4(addr: Array<u8>);
+	case Inet6(addr: Array<u8>);
+}
+
+class SocketBase {
+	// IP address of socket
+	def addr: InetAddress;
+	// Associated file descriptor of socket
+	var fd: int;
```
titzer, line 38: This shouldn't change over the life of the socket, right? If so, probably best to be a `def var`.

rt/x86-64-linux/System.v3
```diff
@@ -8,7 +8,120 @@
...
+	// Server port the socket is connected to
+	def port: int;
+
+	new(addr, localPort, port) {
+		var family: int;
+		match (addr) {
+			Inet4(addrV4) => family = LinuxConst.AF_INET;
+			Inet6(addrV6) => family = LinuxConst.AF_INET6;
+		}
+		fd = System.socket(family, LinuxConst.SOCK_STREAM /* type */, 0 /* protocol */);
```
titzer, line 50: Another thing to think about is auto-closing on finalization of this object. I've started to do that with `RiGc.registerFinalizer`. You have to split the object into a GC'able part and a part (maybe just a closure) that closes the file descriptor.

rt/x86-64-linux/System.v3
k-sareen: Including the system calls like `socket`?

rt/x86-64-linux/System.v3
k-sareen: Correct. Yes, I forgot about `def var`. Thanks

rt/x86-64-linux/System.v3
k-sareen: > You have to split the object into a GC'able part and a part (maybe just a closure) that closes the file descriptor.

Not sure I follow. Would `RiGc.registerFinalizer(sock, sock.close())` not be sufficient? Would you have an example for the same?

rt/x86-64-linux/System.v3
k-sareen: Done. Thanks!

k-sareen:
The `x86-linux` backend fails because the `lib/util/*.v3` files are added to the `-rt.files=`. How am I supposed to add the `DataWriter` dependency? I guess once it's in `lib/net/x86-64-linux`, it should be fine.

k-sareen:
Still not sure how to deal with repeated type definitions for x86_64-linux. Also, apologies, it seems like the x86-linux and x86-64-linux CI configs were swapped. See #118.

rt/x86-64-linux/System.v3
k-sareen: I've moved all the syscalls to `lib/net/x86-64-linux/Net.v3`.

k-sareen:
I've added the `ServerSocket` type now. The `server.accept()` function returns another socket for communication. Here's a sample program using this:

```
def main() {
        def any = Net.ANY_V4;
        def sock = ServerSocket.new(any, 5555);
        def buffer = Array<byte>.new(512);
        System.puts("Hello ");
        System.puti(sock.fd);
        System.ln();
        def ret = sock.bind();
        if (ret == -1) {
                System.error("Error", "Socket connection failed!");
        }
        sock.listen();
        while (true) {
                var connection = sock.accept();
                if (connection.fd == -1) {
                        System.error("Error", "Could not accept connections on socket!");
                }
                System.puts("Connection on fd ");
                System.puti(connection.fd);
                System.ln();
                connection.read(buffer);
                System.puts("recv msg: \"");
                System.puts(buffer);
                System.puts("\"\n");
                connection.write("Hello stranger!\n");
                connection.close();
        }
        sock.close();
}
```

You can use something like `ncat 127.0.0.1 5555 -v` or even just the `ClientSocket` sample program above (I've edited it to be up-to-date). I have to look into how to support a persistent socket connection so one can send multiple messages in the same session (currently that doesn't work as you will see if you run the above `ncat` command and try to send multiple messages).

k-sareen:
By the way, is it possible to have private types or classes? I don't want to make the `IntPointer` type public, but can't figure out a way to make it private. Also, I personally find not being able to set a `def var` variable from a subclass a bit non-intuitive. I'm getting around it with a setter-method, but I think it should be allowed generally without a setter-method.

k-sareen:
Alright -- I've cleaned up a decent amount. I've made it so that you have to manually add the `lib/net` library if you want to use it, instead of it being always there. This allows the CI to pass. I think it looks pretty good. I'll add some more comments etc. tomorrow. Let me know if it looks good.

bin/v3c-x86-64-linux
```diff
@@ -4,5 +4,5 @@ BIN=$(cd $BIN && pwd)
 RT=$(cd $BIN/../rt/ && pwd)
 LIB=$(cd $BIN/../lib/ && pwd)
 V3C=${V3C:=$BIN/v3c}
-RT_FILES=$(echo $RT/x86-64-linux/*.v3 $RT/native/*.v3 $RT/gc/*.v3 $LIB/util/*.v3)
-exec $V3C -heap-size=200m -stack-size=2m -target=x86-64-linux -rt.sttables -rt.gc -rt.gctables -rt.files="$RT_FILES" "$@"
+RT_FILES=$(echo $RT/x86-64-linux/*.v3 $RT/native/*.v3 $RT/gc/*.v3 $LIB/net/x86-64-linux/*.v3)
```
titzer: I think I'd prefer to keep the library files out of the default compile command.

apps/EchoServerClient/DEPS
```diff
@@ -0,0 +1,2 @@
+../../lib/util/*.v3
+../../lib/net/*/*.v3
```
titzer, line 2: Probably want `lib/net/x86-64-linux`; otherwise this will break when new platforms are added.

lib/net/x86-64-linux/Net.v3
```diff
@@ -0,0 +1,267 @@
...
+			// struct sockaddr_in {
+			//   short int          sin_family;  // Address family, AF_INET
+			//   unsigned short int sin_port;    // Port number
+			//   struct in_addr     sin_addr;    // Internet address
+			//   unsigned char      sin_zero[8]; // Same size as struct sockaddr
+			// };
+			def sockAddrBldr = DataWriter.new();
+			def port = if(server, socket.localPort, socket.port);
+			sockAddrBldr.acquire(LinuxConst.SOCKADDR_V4_SIZE);
+			def addrInt = (int.view(addrV4[0]) << 24)
```
titzer, line 94: Hmm, if the address is already a byte array, maybe we can get away with just copying the bytes of it over, instead of packing it into an `int` and then calling `put_b32be`?

lib/net/x86-64-linux/Net.v3
```diff
@@ -0,0 +1,267 @@
...
+// types, Inet4Array and Inet6Array, like so:
+// class Inet4Array {
+// 	def addr: Array<u8> = Array.new(4);
+// }
+// class Inet6Array {
+// 	def addr: Array<u8> = Array.new(16);
+// }
+// type InetAddress {
+// 	case Inet4(addr: Inet4Array);
+// 	case Inet6(addr: Inet6Array);
```
titzer, line 18: I think it's best to leave Inet6 out until it's fully implemented.

titzer (changes requested):

bin/v3c-x86-64-linux
k-sareen: I've removed them yes.

lib/net/x86-64-linux/Net.v3
k-sareen: Sure. Is it fine if I just comment it out instead of deleting it? There's some substantial stuff for IPv6 like the rendering code.

lib/net/x86-64-linux/Net.v3
k-sareen: Done.

lib/net/x86-64-linux/Net.v3
k-sareen: PS do you have a suggestion on how to deal with array sizes for the two kinds of IP addresses? Ideally the arrays are only 4- and 16-bytes long for IPv4 and IPv6 respectively. Currently it's just `Array<u8>` which can be any array length.

titzer: I think it's best not to have commented-out code checked in. I only very rarely do. I keep experimental code in a separate directory or as a sketch outside the repo.

titzer: For the array sizes, what about this?

```
type INetAddress {
    case IPV4(oct3: u8, oct2: u8, oct1: u8, oct0: u8) { }
    case IPV6(word3: u32, word2: u32, word1: u32, word9: u32) { }

    def put(r: Range<u8>) {
        match (this) {
           case IPV4(b3, b2, b1, b0) => { r[0] = b3; r[1] = b2; r[2] = b1; r[2] = b1; r[3] = b0; }
           case IPV6 ...
        }
    }
```

That way, `INetAddress` values will always compare equal, and there will be no array object overhead.

k-sareen: It works for IPv4 but I feel like it'll make some operations for IPv6 a bit painful such as the rendering code. It's nicer to work with bytes directly instead of having to pack and unpack them into a `u32`.

> Also, using a Range<T> allows writing the types out without relying on DataReader.

Not sure I follow. I'm not using a `DataReader` anywhere in the code, just directly copying to/from the C-struct definition.

k-sareen: I added IPv6 support -- it was surprisingly easy to add. I'll think about how to fix up the array size issue.

lib/net/x86-64-linux/Net.v3
```diff
@@ -0,0 +1,371 @@
+// Copyright 2022 Kunal Sareen. All rights reserved.
```
titzer, line 1: Can you make this copyright the Virgil authors?

lib/net/x86-64-linux/Net.v3
```diff
@@ -0,0 +1,371 @@
...
+		} else {
+			fd = sockfd;
+		}
+	}
+
+	def setLocalPort(local: int) {
+		localPort = local;
+	}
+
+	/**
```
titzer, line 177: Can you use `//` comments? I tend to prefer saving lines.

titzer:
I think after the last two minor changes are addressed, we can land this. Thanks Kunal!

titzer (changes requested):

k-sareen:
> I think after the last two minor changes are addressed, we can land this. Thanks Kunal!

Cool! I've also made the changes regarding the array sizes. I'll push it all in one go.

lib/net/x86-64-linux/Net.v3
k-sareen: Done.

lib/net/x86-64-linux/Net.v3
k-sareen: Done.

k-sareen:
> Cool! I've also made the changes regarding the array sizes. I'll push it all in one go.

I've made the function `rangeToInetAddress()` public so that you can easily use it for making your own IP addresses instead of hand-crafting it. As an aside, it'd be good to add tests for it (I've on my own machine tested the generated struct layout and it exactly matches the struct layout from an equivalent C program), but we can figure that out in a future PR.

Merry Christmas! 🎉

k-sareen:
This is super cool! I just ran the `EchoServer` on one computer and the `EchoClient` on another and they were able to talk to each other. Pretty fun. UDP is probably the next thing I'll add.

titzer:
Thanks Kunal!

PR #120, opened 2022-12-28
[lib/net]: Add UDP support and make error checking robust

k-sareen:
This commit makes the error checking and handling more robust and adds support for connectionless and connected UDP sockets.


k-sareen:
I was being a bit miserly in terms of creating garbage so I added the `packet` field to `recvfrom()` even though it's not strictly necessary. If you provide a packet, it will override its values instead of allocating a new object.

apps/EchoServerClient/TcpEchoServer.v3
```diff
@@ -30,16 +30,6 @@ def main(args: Array<string>) {
 	strBldr.put1("Hello! Socket opened on fd %d\n", sock.fd);
```
titzer, line 30: I usually use the variable name `buf` for `StringBuilder`--it used to be called `StringBuffer`, so the `buf` name was already pervasive.

apps/EchoServerClient/UdpEchoClient.v3
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	def buffer = Array<byte>.new(512);
+	def strBldr = StringBuilder.new();
+
+	if (args.length == 0 || args.length > 1) {
+		usage();
+	}
+
+	strBldr.put1("Hello! Socket opened on fd %d\n", sock.fd);
```
titzer, line 45: Here too, but since there is already a `buffer`, I might call the byte array the `iobuf` or something.

lib/net/x86-64-linux/Net.v3
```diff
@@ -211,37 +378,74 @@ class ClientSocket extends Socket {
...
-	// Associated sockaddr structure. Is instantiated in bind()
+	// Associated sockaddr structure. Is instantiated in constructor
 	private var sockAddr: Array<byte>;
 
-	new(addr: InetAddress, localPort: int) super(addr, localPort, -1 /* port */, -1 /* fd */, true) { }
+	new(addr: InetAddress, localPort: int) super(addr, -1 /* port */, localPort, SocketType.Tcp, -1 /* fd */, true) {
+		sockAddr = toSockAddr(addr, localPort);
+		var ret = bind();
+		if (ret == -1) {
+			System.error("Error", "Could not bind socket!");
```
titzer, line 399: `System.error` brings down the process, which is pretty violent. These error conditions should be encoded as return values to allow the surrounding program to do something. Thus, this functionality shouldn't really be in the constructor. Same for base behavior. Constructors should be to initialize the object, and possibly-failing operations need to be separate.

I haven't thought about this deeply for networking, but one way would be move some of the functionality into component methods (e.g. helpers that create these objects (or `null` if failure). I'm not great at API design, so take that with a grain of salt.

lib/net/x86-64-linux/Net.v3
```diff
@@ -327,6 +537,45 @@ component Net {
...
+			LinuxConst.SYS_sendto, (
+				sockfd,
+				Pointer.atContents(msg),
+				msg.length,
+				0 /* flags */,
+				Pointer.atContents(sockAddr),
+				sockAddr.length
+			)
+		);
+		return if(ret.0 >= 0, int.view(ret.0), -1);
```
titzer, line 556: Another thing to think about is how to expose finer-granularity error conditions to applications using the library. I cut a lot of corners in making `System` very very small, but a networking library should give decent errors to help troubleshoot applications.

titzer (changes requested):

lib/net/x86-64-linux/Net.v3
k-sareen: Admittedly `System.error` is quite spartan. I was trying to make the library a bit more hands-free wherein you don't have to manually `bind()` then `listen()` etc etc. I'm thinking maybe a new method like `open()` (or `start()`?) that will perform all the actions.

With regard to errors (in a different comment, but related to this), I can make something like a result type (`type NetResult<T> { Result(v: T); Error(e: NetError); }`; where `NetError` is some `enum` of possible errors). Does that sound better?

lib/net/x86-64-linux/Net.v3
k-sareen: Right. That's fair. I was emulating the `System` component in-terms of the syscalls, but yeah there is room for providing better errors. We should add an `enum` for `errno` values and their respective error strings. It would be useful imo.

lib/net/x86-64-linux/Net.v3
```diff
@@ -150,59 +147,229 @@ def socketToSockAddr(socket: Socket, server: bool) -> Array<byte> {
...
-	// Write a message {msg} into the socket.
-	def write(msg: string) {
-		System.fileWriteK(fd, msg, 0, msg.length);
+	// Receive a message into buffer {buf} from a remote host. Users can provide
+	// an instance of {packet} which will be overridden with information about
+	// the remote host. If no {packet} is provided, a new one is allocated and
+	// returned. Returns the number of bytes read into {buf} and information
+	// about the remote host for future communication. The receive will fail for
+	// a TCP socket.
+	def recvfrom(buf: Array<byte>, packet: UdpPacket) -> (int, UdpPacket) {
```
titzer, line 244: If you accept `Range<byte>` here, users can pass sub-ranges of arrays (and `Array` auto-promotes to `Range`).

I'm thinking of going through most/all APIs that take arrays and upgrading them to ranges.

lib/net/x86-64-linux/Net.v3
```diff
@@ -150,59 +147,229 @@ def socketToSockAddr(socket: Socket, server: bool) -> Array<byte> {
...
+				System.error("Error", "Unexpected size for returned sockaddr!");
+			}
+		}
+
+		return (-1, null);
+	}
+
+	// Send a message {msg} into the socket. The send will fail on a standard UDP
+	// connectionless socket.
+	def send(msg: string) {
```
titzer, line 283: Same here, I think `Range<byte>` would be more general.

I also have plans of making `string` into a proper immutable array of Unicode code points.

lib/net/x86-64-linux/Net.v3
titzer: Yeah, that sounds alright for now. At some point when more libraries accumulate we'll know better what looks good.

lib/net/x86-64-linux/Net.v3
k-sareen: Yeah I noticed this behaviour with the `rangeToInetAddress` function. It's very cool! Good idea to change it to a `Range` type

lib/net/x86-64-linux/Net.v3
k-sareen: Should the call to `Net.socket()` in the `Socket` constructor also be moved? It sets the `fd` field, so I'm of the opinion it belongs in the constructor. But it is a syscall and it can fail. I'm leaning towards keeping it in the `Socket` constructor unless you have an opinion on this.

mosheduminer: Just give drop a comment here.

Although it feels really weird from an API perspective, my two cents is that it shouldn't be in the constructor, because there's no way to recover / indicate failure to the caller from a constructor. Maybe if there was a better error handling story you could hook into that instead, but right now the only way to inform the caller of errors is through a return value.

That said, it's quite unlikely this would fail...

k-sareen: Yeah going by the error return values for the `socket()` syscall, it is unlikely to fail for the (hardcoded) values we have. But I wanted to check if it is considered okay to use it in the constructor, or maybe just add it to the `open()` method that I'll make.

k-sareen:
Sorry for being AWOL. Been really busy with work. Two questions:
 1. How would I convert an `int` directly into the `enum` variant that it corresponds to? Or on a higher-level, how would I convert between one `enum` (Linux errnos) to a more descriptive and general `enum` (for Network errors) where the conversion should be a bit shift operation between tags.
 2. If instead of returning an `IntPointer` type (which is now private), I now return the actual `int` field. My question is the compiler won't optimize the pointer load right to a (potentially) stale register value, right? C uses the `volatile` keyword to prevent this, but I'm unsure how Virgil manages this.

k-sareen:
@titzer Couple of things:

1. Using the `NetResult` type is _extremely_ verbose. Main issues are: (i) Have to use unqualified case (i.e. `x: Case`) as case-expression has type parameter; and (ii) Have to convert from `NetResult<T>` to `NetResult<Y>` even though `NetResult.Error` is the same type (I've added comments in `Socket.open()` which mention this).
2. I couldn't change the API to use `Range<byte>` as `Ptr.atContents` and a bunch of other functions (such as `System.fileWriteK`) are only defined for `Array<byte>`.
3. I'm trusting the compiler won't optimize a pointer load with a potentially stale value with the `IntPointer` class. See `Net.accept` for an example.

k-sareen:
@titzer if you had time, could you review this PR as well? Sorry for being AWOL, got busy with work.

titzer:
> 1. Using the `NetResult` type is _extremely_ verbose. Main issues are: (i) Have to use unqualified case (i.e. `x: Case`) as case-expression has type parameter; and (ii) Have to convert from `NetResult<T>` to `NetResult<Y>` even though `NetResult.Error` is the same type (I've added comments in `Socket.open()` which mention this).

I hear ya, I see it, and it does not look pretty :-)

One suggestion: you could add a method to `NetResult`:

```
type NetResult<T> {
  case Ok(val: T);
  case Error(msg: string);
  def force(f: string -> void) -> T {
    match (this) {
      Ok(val) => return val;
      Error(msg) => f(msg);
    }
    var d: T;
    return d;
  }
}

def fatal = System.error("SocketError", _);
var x = someSocketCall().force(fatal); // unpacks upon success, calls {fatal} otherwise
```

> 2. I couldn't change the API to use `Range<byte>` as `Ptr.atContents` and a bunch of other functions (such as `System.fileWriteK`) are only defined for `Array<byte>`. I think changing everything to accept a `Range` type is out-of-scope of this PR. We can figure it out in the future.

Ah, true. `Pointer.atContents` will have to be upgraded.

> 3. I'm trusting the compiler won't optimize a pointer load to a potentially stale (register) value with the `IntPointer` class. See `Net.accept` for an example.

That part still feels clunky. Allocating an object to store just an `int` (and passing a pointer to the field, as opposed to a pointer into an array) is somewhat dangerous; that will need pinning too. But the answer is that the compiler backs off optimizations for fields that have had their address taken with `Pointer.atField`.

> 4. I've made the API such that a user has to call `socket.open()` _before_ they perform any socket operation.
> 5. In general it gets pretty tiring to write every type/enum variant fully qualified. Is there a way to directly use, for example,`InvalidInput` instead of `NetError.InvalidInput`?

There is a `match` statement variant that unpacks the cases; see above.

k-sareen:
> > 1. Using the `NetResult` type is _extremely_ verbose. Main issues are: (i) Have to use unqualified case (i.e. `x: Case`) as case-expression has type parameter; and (ii) Have to convert from `NetResult<T>` to `NetResult<Y>` even though `NetResult.Error` is the same type (I've added comments in `Socket.open()` which mention this).
> 
> I hear ya, I see it, and it does not look pretty :-)
> 
> One suggestion: you could add a method to `NetResult`:
> 
> ```
> type NetResult<T> {
>   case Ok(val: T);
>   case Error(msg: string);
>   def force(f: string -> void) -> T {
>     match (this) {
>       Ok(val) => return val;
>       Error(msg) => f(msg);
>     }
>     var d: T;
>     return d;
>   }
> }
> 
> def fatal = System.error("SocketError", _);
> var x = someSocketCall().force(fatal); // unpacks upon success, calls {fatal} otherwise
> ```

Ah yes. I was thinking an equivalent to Rust's `.expect()`/`.unwrap()` would be useful too. Yeah it definitely looks cleaner. Will refactor it.

> > 3. I'm trusting the compiler won't optimize a pointer load to a potentially stale (register) value with the `IntPointer` class. See `Net.accept` for an example.
> 
> That part still feels clunky. Allocating an object to store just an `int` (and passing a pointer to the field, as opposed to a pointer into an array) is somewhat dangerous; that will need pinning too. But the answer is that the compiler backs off optimizations for fields that have had their address taken with `Pointer.atField`.

What would the difference be between an `Array<int>` and an `class IntPointer`? From the perspective of the GC, they should both be pinned and on the heap, no? 
 
> > 4. I've made the API such that a user has to call `socket.open()` _before_ they perform any socket operation.
> > 5. In general it gets pretty tiring to write every type/enum variant fully qualified. Is there a way to directly use, for example,`InvalidInput` instead of `NetError.InvalidInput`?
> 
> There is a `match` statement variant that unpacks the cases; see above.

Ah I see. So the cases are only inferred in a `match (this)`? What about when I just want to refer to a specific variant of an `enum` without fully qualifying it, as with `NetError.*`?

Also are you happy with the current definition of `NetResult` or should I adapt it to be similar to what you have in your example where the `Error` variant has as `string` field instead of a `NetError` field.

titzer:
> What would the difference be between an `Array<int>` and an `class IntPointer`? From the perspective of the GC, they should both be pinned and on the heap, no?
> 

It's just more code to define a new class. The right solution in the end will be to define a `struct` and use it as a view over an `Array<byte>`. I'd prefer an `Array<int>` if for no other reason than it's shorter.

> > > 4. I've made the API such that a user has to call `socket.open()` _before_ they perform any socket operation.
> > > 5. In general it gets pretty tiring to write every type/enum variant fully qualified. Is there a way to directly use, for example,`InvalidInput` instead of `NetError.InvalidInput`?
> > 
> > 
> > There is a `match` statement variant that unpacks the cases; see above.
> 
> Ah I see. So the cases are only inferred in a `match (this)`? What about when I just want to refer to a specific variant of an `enum` without fully qualifying it, as with `NetError.*`?

It doesn't need to be `this`, any expression that is of a variant type can be matched over with deconstruction.

> Also are you happy with the current definition of `NetResult` or should I adapt it to be similar to what you have in your example where the `Error` variant has as `string` field instead of a `NetError` field.

I think it's OK to land now; sorry I've been so pedantic. At some point I'll write some code against the API to get a better feel for it.

k-sareen:
> > What would the difference be between an `Array<int>` and an `class IntPointer`? From the perspective of the GC, they should both be pinned and on the heap, no?
> 
> It's just more code to define a new class. The right solution in the end will be to define a `struct` and use it as a view over an `Array<byte>`. I'd prefer an `Array<int>` if for no other reason than it's shorter.

Right -- that makes sense. I'll change it to an `Array<int>`.
 
> > Ah I see. So the cases are only inferred in a `match (this)`? What about when I just want to refer to a specific variant of an `enum` without fully qualifying it, as with `NetError.*`?
> 
> It doesn't need to be `this`, any expression that is of a variant type can be matched over with deconstruction.

Sorry, not sure I follow? I had previously tried to do that with `NetResult` in some functions and the compiler gave me an error about having to use an unqualified case. I guess that error has to do more with the type parameter?

> > Also are you happy with the current definition of `NetResult` or should I adapt it to be similar to what you have in your example where the `Error` variant has as `string` field instead of a `NetError` field.
> 
> I think it's OK to land now; sorry I've been so pedantic. At some point I'll write some code against the API to get a better feel for it.

Yeah I'm not too happy with the UX of the API right now, but the majority of the issue is just the verbosity, so maybe it'll be fixed when I add the `.force()` function/when the `?` operator is added for error propagation.

k-sareen:
@titzer So I've added the above API, which definitely cleans up the user-facing API significantly. However, the library internals are still very messy due to the aforementioned mentioned reasons:

> Using the NetResult type is extremely verbose. Main issues are: (i) Have to use unqualified case (i.e. x: Case) as case-expression has type parameter; and (ii) Have to convert from NetResult\<T> to NetResult\<Y> even though NetResult.Error is the same type (I've added comments in Socket.open() which mention this).

I think the type system should be able to deal with the second issue as there is no real reason the conversion is not possible. Not sure how to deal with the first issue. Technically the generic parameter `T` should be easily inferred since it's in the return type. So maybe the type system just has to be updated to be able to infer the generic type there.

PR #123, opened 2022-12-29
[struct] implemented lookup struct, subtype, nested strcutdecl etc.

refusal-type-bamboo:
Hi Dr.Titzer,

All tests except for 2 (repeat13.v3 and offset00.v3) pass. I also modified some test cases, let me know if I made a mistake

Best
Yunqing


aeneas/src/v3/TypeSystem.v3
```diff
@@ -59,6 +59,18 @@ component TypeSystem {
 				}
 				return false;
 			}
+			V3Kind.REF => {
+				if (t1 == Null.TYPE) return false;
+				if (t1.typeCon.kind != V3Kind.REF) return false;
+				var reft1 = V3Ref_TypeCon.!(t1.typeCon);
```
titzer, line 65: The more robust way to do this is to implement `getSuperType()` for struct ref types. That should return `ref<T>` for `Ref<T>` types and `null` for everything else. Then use a similar style of loop from the CLASS case above.

aeneas/src/core/Operator.v3
```diff
@@ -339,6 +339,7 @@ component V3Op {
 			V3Kind.CLASS,
 			V3Kind.FUNCREF => opcode = Opcode.RefEq;
 			V3Kind.VARIANT => opcode = Opcode.VariantEq;
+			V3Kind.REF => opcode = Opcode.RefEq;
```
titzer, line 342: Sorry, this is a bit confusing here, but the `RefEq` operator works on "pointer" references, which are a single word. It's probably fine to leave this as an overloaded equal for now, because it will become a comparison of the byte array and offset in the middle of the compiler (similar to how ranges work).

aeneas/src/v3/TypeSystem.v3
```diff
@@ -123,6 +135,9 @@ component TypeSystem {
 			return null;
 		} else if (V3.isTuple(t1) && V3.isTuple(t2)) {
 			return mergeLists(t1.nested, t2.nested, Tuple.TYPECON, upper);
+		} else if (V3.isRef(t1) && V3.isRef(t2)) {
```
titzer, line 138: You probably also want to implement a case for `isNullable()`, which will hit above.

aeneas/src/vst/MethodEnv.v3
```diff
@@ -871,6 +891,13 @@ class MethodEnv {
 			if (ttype == Void.TYPE) return expr.bind(null, VarBinding.Void, ttype);
 			return expr.bind(null, VarBinding.Type(ttype), ttype);
 		}
+		// lookup a struct
+		var structName = expr.ident.name.image;
+		for (i < verifier.file.structs.length) {
```
titzer, line 896: There is a `HashMap` in the program so you can do one hashtable lookup here.

aeneas/src/vst/Verifier.v3
```diff
@@ -195,6 +195,41 @@ class Verifier(compiler: Compiler, prog: Program) {
...
+	def checkFieldCycle(decl: VstStruct) {
+		var cv = decl.verifier;
+		for (l = decl.members; l != null; l = l.tail) {
+			var buffer = StringBuilder.new();
+			VstStructField.!(l.head).mtref.render(buffer);
+			var structName = buffer.toString();
+			for (i < cv.file.structs.length) {
+				if (Strings.equal(structName, cv.file.structs[i].name())){
+					if (checkStructCycle(cv.file.structs[i])) {
+						cv.FieldDeclError(l.head, "cyclic field declaration detected");
```
titzer, line 207: Can you inline the error case into the recursive call (i.e. not return a boolean)?

Also, please report the members of the cycle; see the cyclic inheritance check.

aeneas/src/vst/MethodEnv.v3
```diff
@@ -145,6 +145,26 @@ class MethodEnv {
 		}
 		return VarBinding.None;
 	}
+	def lookupStructTypeMember(expr: VarExpr, structName: VstStruct) -> VarBinding {
```
titzer, line 148: Can we name this lookupStructMember? (because structs aren't really types).

aeneas/src/vst/Verifier.v3
```diff
@@ -1687,6 +1726,11 @@ class TypeChecker(ERROR: ErrorGen, file: VstFile) extends VstVisitor<Type, Type>
 			else UnresolvedOp(expr.op.token, ltype, null);
 			return getErrorType();
 		}
+		if (op.opcode == Opcode.RefEq && !TypeSystem.isSubtype(ltype, rtype) &&
```
titzer, line 1729: I don't think you need this extra case; it should handle incomparable types by default already.

titzer (changes requested):
Please don't commit `bin/v3c`; it overwrites itself when it detects your platform, and you'd be committing one specific configuration.

Please do:

```
% git checkout HEAD bin/v3c
% git update-index --assume-unchanged bin/v3c
```

aeneas/src/vst/MethodEnv.v3
refusal-type-bamboo: It seems that MethodEnv don't have access to the Program, so should I leave this be?

aeneas/src/vst/Verifier.v3
refusal-type-bamboo: If so, a `TypeError` instead of a `Comparison Error `is generated in test case `cmp02.v3`. Should I change the test case?

titzer: Yeah, please update the test case.

aeneas/src/vst/MethodEnv.v3
titzer: You can get to the program through `verifier.verifier.prog`.

refusal-type-bamboo: thank you! Lookup now uses the hashmap

aeneas/src/vst/Verifier.v3
refusal-type-bamboo: test case updated

PR #124, opened 2022-12-30
[struct] Changes made

refusal-type-bamboo:


aeneas/src/core/Operator.v3
```diff
@@ -339,6 +339,7 @@ component V3Op {
 			V3Kind.CLASS,
 			V3Kind.FUNCREF => opcode = Opcode.RefEq;
 			V3Kind.VARIANT => opcode = Opcode.VariantEq;
+			V3Kind.REF => opcode = Opcode.RefEq;
```
titzer, line 342: I think you should remove this case.

titzer:
Please see my comment on the previous PR about `bin/v3c`, you should not commit changes to this file.

aeneas/src/vst/Vst.v3
```diff
@@ -858,6 +858,7 @@ type VarBinding {
 	case Apply(op: Operator, facts: Fact.set);		// apply {op} to receiver
 	case Partial(op: Operator, facts: Fact.set);		// partially apply {op} to receiver
 	case PtrAtField(ptrType: Type);
+	case StructName(structName: VstStruct);
```
titzer, line 861: There should also be a binding for a struct name "Struct.field.offset"?

aeneas/src/vst/MethodEnv.v3
```diff
@@ -145,6 +145,26 @@ class MethodEnv {
...
 	}
+	def lookupStructMember(expr: VarExpr, structName: VstStruct) -> VarBinding {
+		var name = expr.ident.name;
+		if (Strings.equal("size", name.image)) {
+			// found struct size
+			def etype = IntType.new(32, true, Int_TypeCon.new(32, true));
+			def size = etype.box(structName.byteSize);
+			return expr.bind(expr.expr, VarBinding.Const(size, etype), etype);
+		}
+		for (l = structName.members; l != null; l = l.tail) {
```
titzer, line 156: There is a `memberMap` which is a hashmap to avoid a linear lookup.

aeneas/src/vst/Verifier.v3
```diff
@@ -197,6 +197,22 @@ class Verifier(compiler: Compiler, prog: Program) {
 			}
 		}
 	}
+	def checkStructCycle(decl: VstStruct){
+		var cv = decl.verifier;
+		if (cv.onstack) {
+			cv.FieldDeclError2(decl.token.range(), "cyclic field declaration detected");
+		} else {
+			cv.onstack = true;
+			for (l = decl.members; l != null; l = l.tail) {
```
titzer, line 206: There is a member map, and the `mtref` should have a string; you shouldn't need to render it.

titzer (commented):
Looks good with minor revisions.

aeneas/src/vst/Verifier.v3
refusal-type-bamboo: I now uses the string in `mtref`. However, I think looping through all members is necessary since here I'm not searching for a member but checking whether any of the members will result in a cycle.

titzer (approved):
lgtm!

PR #126, opened 2023-01-15
[exp] struct normalization added

refusal-type-bamboo:
Hi Dr.Titzer,

All tests except for write_enum passes, let me know if there's any questions.

Best
Yunqing

### DIFF
--- a/aeneas/src/ir/SsaNormalizer.v3
+++ b/aeneas/src/ir/SsaNormalizer.v3
@@ -537,8 +537,8 @@ class SsaRaNormalizer extends SsaRebuilder {
 		}
 	}
 	def genGetBytesInstr(app: SsaApplyOp, array: SsaInstr, pos: SsaInstr, offset: int, width: int,
-		// TODO: little-endian only for now
 		intType: IntType, result: SsaInstr) -> SsaInstr {
+		// TODO: little-endian only for now
 		var an = ArrayNorm.!(normType(V3.arrayByteType));
 		for (i < width) {
 			var read = curBlock.opArrayGetElem(an.newType, Int.TYPE, app.facts,
@@ -551,9 +551,8 @@ class SsaRaNormalizer extends SsaRebuilder {
 		return result;
 	}
 	def genSetBytesInstr(app: SsaApplyOp, val: SsaInstr, width: int, array: SsaInstr, pos: SsaInstr,
+		offset: int, intType: IntType) {
 		// TODO: little-endian only for now
-		offset: int, 	intType: IntType) {
-		var results = Array<SsaInstr>.new(width);
 		def mask = newGraph.intConst(255);
 		for (i < width) {
 			var shifted = curBlock.addApply(app.source, intType.opShr(),
@@ -562,7 +561,6 @@ class SsaRaNormalizer extends SsaRebuilder {
 				value = curBlock.opIntViewI0(value.getType(), Byte.TYPE, value);
 			var write = curBlock.opArraySetElem(array.getType(), Int.TYPE, app.facts,
 				array, curBlock.opIntAdd(newGraph.intConst(i + offset), pos), value);
-			results[i] = write;
 		}
 	}
 	def normDefault(app: SsaApplyOp, op: Operator) {
--- a/aeneas/src/ir/SsaNormalizer.v3
+++ b/aeneas/src/ir/SsaNormalizer.v3
@@ -377,30 +377,16 @@ class SsaRaNormalizer extends SsaRebuilder {
 				var result = SsaInstr.!(newGraph.intConst(0));
 				match (ortype) {
 					x: IntType => {
-						var width = IntType.!(ortype).packedByteSize();
-						// assume little endianess
-						for (i < width) {
-							var read = curBlock.opArrayGetElem(an.newType, Int.TYPE, app.facts,
-								newArgs[0], curBlock.opIntAdd(newGraph.intConst(i + offset), pos));
-							var shifted = curBlock.addApply(app.source, IntType.!(ortype).opShl(),
-								[read, newGraph.intConst(8 * i)]);
-							result = curBlock.addApply(app.source, IntType.!(ortype).opOr(),
-								[shifted, result]);
-						}
+						var width = x.packedByteSize();
+						result = genGetBytesInstr(app, array, pos,
+							offset, width, true, x, result);
 						return map1(app, result);
 					}
 					x: EnumType => {
-						var width = EnumType.!(ortype).packedByteSize();
-						for (i < width) {
-							var read = curBlock.opArrayGetElem(an.newType, Int.TYPE, app.facts,
-								newArgs[0], curBlock.opIntAdd(newGraph.intConst(i + offset), pos));
-							var longViewOp = V3Op.newIntViewI(read.getType(), Long.TYPE);
-							read = curBlock.addApply(app.source, longViewOp, [read]);
-							var shifted = curBlock.addApply(app.source, Long.TYPE.opShl(),
-								[read, newGraph.intConst(8 * i)]);
-							result = curBlock.addApply(app.source, Long.TYPE.opOr(),
-								[shifted, result]);
-						}
+						var width = x.packedByteSize();
+						var enumType = x.enumDecl.tagType;
+						result = genGetBytesInstr(app, array, pos,
+							offset, width, true, Int.TYPE, result);
 						var enumTagLen = newGraph.intConst(x.enumDecl.cases.length);
 						var inBound = curBlock.opIntULt(IntType.!(result.getType()),
 							Int.TYPE, result, enumTagLen);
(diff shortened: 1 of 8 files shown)


aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -346,6 +346,306 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+								[newArgs[2], newGraph.intConst(8 * i)]);
+							var value = curBlock.opIntAnd(shifted, mask);
+							value = curBlock.opIntViewI0(value.getType(), Byte.TYPE, value);
+							var write = curBlock.opArraySetElem(array.getType(), Byte.TYPE, app.facts,
+								newArgs[0], curBlock.opIntAdd(newGraph.intConst(i + offset), pos), value);
+							results[i] = write;
+						}
+						return mapN(app, results);
+					}
+					/* x: EnumType => {
```
titzer, line 461: This code is still commented out. It should be very similar to the `int` case, except you might need to do a conversion from the enum type to an integer type first.

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -346,6 +346,306 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+						for (i < width) {
+							var shifted = curBlock.addApply(app.source, IntType.!(x).opShr(),
+								[newArgs[2], newGraph.intConst(8 * i)]);
+							var value = curBlock.opIntAnd(shifted, mask);
+							value = curBlock.opIntViewI0(value.getType(), Byte.TYPE, value);
+							var write = curBlock.opArraySetElem(array.getType(), Byte.TYPE, app.facts,
+								newArgs[0], curBlock.opIntAdd(newGraph.intConst(i + offset), pos), value);
+							results[i] = write;
+						}
+						return mapN(app, results);
```
titzer, line 459: You don't need to `mapN()` the result of stores.

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -346,6 +346,306 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+				var newArgs = genRefs(args);
+				var array = newArgs[0], pos = newArgs[1];
+				match(newArgs[2].getType()) {
+					x: IntType => {
+						var width = IntType.!(x).byteSize();
+						//Terminal.put1("%d\n", width);
+						var results = Array<SsaInstr>.new(width);
+						def mask = newGraph.intConst(255);
+						for (i < width) {
+							var shifted = curBlock.addApply(app.source, IntType.!(x).opShr(),
```
titzer, line 451: No need to cast `IntType.!(x)`

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -346,6 +346,306 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+								[newArgs[2], newGraph.intConst(8 * i)]);
+							var value = curBlock.opIntAnd(shifted, mask);
+							value = curBlock.opIntViewI0(value.getType(), Byte.TYPE, value);
+							var write = curBlock.opArraySetElem(array.getType(), Byte.TYPE, app.facts,
```
titzer, line 455: The second argument to `opArraySetElem` is the type of the index (i.e. `int`), not `byte`.

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -346,6 +346,306 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+						curBlock = split.addIf(inBound);
+						curBlock = split.addElse();
+						curBlock = split.finish();
+						result = split.addPhi(Long.TYPE, [result, newGraph.longConst(0)]);
+						var intViewOp = V3Op.newIntViewI(result.getType(), nrtype);
+						result = curBlock.addApply(app.source, intViewOp, [result]);
+						return map1(app, result);
+					}
+					x: FloatType => {
+						var width = FloatType.!(ortype).byteSize();
```
titzer, line 417: There's no need to cast `x` here (or elsewhere).

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -346,6 +346,306 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+								[shifted, result]);
+						}
+						return map1(app, result);
+					}
+					x: EnumType => {
+						var width = EnumType.!(ortype).packedByteSize();
+						for (i < width) {
+							var read = curBlock.opArrayGetElem(an.newType, Int.TYPE, app.facts,
+								newArgs[0], curBlock.opIntAdd(newGraph.intConst(i + offset), pos));
+							var longViewOp = V3Op.newIntViewI(read.getType(), Long.TYPE);
```
titzer, line 397: Instead of Long.TYPE, I think you want to use the enum's tag type, either from `V3.getVariantTagType` or from `x.enumDecl.tagType`.

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -346,6 +346,306 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+				var an = ArrayNorm.!(normType(V3.arrayByteType));
+				var ortype = normReturnType(op).oldType, nrtype = normReturnType(op).newType;
+				var newArgs = genRefs(app.inputs);
+				var array = newArgs[0], pos = newArgs[an.size];
+				var result = SsaInstr.!(newGraph.intConst(0));
+				match (ortype) {
+					x: IntType => {
+						var width = IntType.!(ortype).packedByteSize();
+						// assume little endianess
+						for (i < width) {
```
titzer, line 382: Can you factor out the read and write loops for the various cases?

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -346,6 +346,306 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+						}
+						return mapN(app, results);
+					}*/
+					x: FloatType => {
+						var width = FloatType.!(x).byteSize();
+						var results = Array<SsaInstr>.new(width);
+						def mask = newGraph.longConst(255);
+						newArgs[2] = curBlock.opIntView(newArgs[2].getType(), Long.TYPE, newArgs[2]);
+						for (i < width) {
+							var shifted = curBlock.addApply(app.source, Long.TYPE.opShr(),
```
titzer, line 483: You'll want to use the integer type corresponding to the float type's width. For example `Int.getType(false, x.total_width)`.

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -346,6 +346,306 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+								[newArgs[2], newGraph.intConst(8 * i)]);
+							var value = curBlock.opIntAnd(shifted, mask);
+							value = curBlock.opIntViewI0(value.getType(), Byte.TYPE, value);
+							var write = curBlock.opArraySetElem(array.getType(), Byte.TYPE, app.facts,
+								newArgs[0], curBlock.opIntAdd(idx, newGraph.intConst(i)), value);
+							results[i] = write;
+						}
+						return mapN(app, results);
+					}
+					x: FloatType => {
```
titzer, line 628: The `FloatType` case is duplicated here.

titzer (changes requested):

aeneas/src/ir/SsaNormalizer.v3
refusal-type-bamboo: I still used Int.TYPE since there might be the case that the value read in is greater than the maximum of tag value and we need to return an oob 0. The type of the tag is usually too short. Suppose we're reading in 0x128 for a enum type with maximum tag value 2, we should be returning 0, but if we use the tag's type, we'll fail to catch this oob.

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -346,6 +346,179 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+						genSetBytesInstr(app, val, width, array, pos, offset, true, Int.TYPE);
+					}
+					x: EnumType => {
+						var width = x.byteSize();
+						val = curBlock.opIntViewI0(newArgs[2].getType(), Int.TYPE, val);
+						genSetBytesInstr(app, val, width, array, pos, offset, true, Int.TYPE);
+					}
+					x: FloatType => {
+						var width = x.byteSize();
+						val = curBlock.opIntView(val.getType(), Long.TYPE, val);
```
titzer, line 436: Here I think we should use `Int.getType(false, x.total_width)` instead of `Long.TYPE`. Also I think you can just use `x` instead of `val.getType()`.

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -364,6 +537,33 @@ class SsaRaNormalizer extends SsaRebuilder {
 			}
 		}
 	}
+	def genGetBytesInstr(app: SsaApplyOp, array: SsaInstr, pos: SsaInstr, offset: int, width: int,
+		small_endianess: bool, intType: IntType, result: SsaInstr) -> SsaInstr {
```
titzer, line 541: I think it's OK to remove the `small_endianess` argument for now and leave:

```
// TODO: little-endian only for now
```

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -364,6 +537,33 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+			var shifted = curBlock.addApply(app.source, intType.opShl(),
+				[read, newGraph.intConst(8 * i)]);
+			result = curBlock.addApply(app.source, intType.opOr(),
+				[shifted, result]);
+		}
+		return result;
+	}
+	def genSetBytesInstr(app: SsaApplyOp, val: SsaInstr, width: int, array: SsaInstr, pos: SsaInstr,
+		offset: int, small_endianess: bool, intType: IntType) {
+		var results = Array<SsaInstr>.new(width);
```
titzer, line 555: `results` are not used; can remove.

titzer (commented):
Just some minor cleanups; can you bump the version number and then I'll merge?

Thanks!

titzer:
Also, please run the `test. bash` script in the `test/struct/` directory. It runs on all the platforms, which show some of the issues that result from mismatches in `int` vs `long`.

refusal-type-bamboo:
Hi Dr.Titzer,

The changes are made accordingly. However, when running test.bash, there're too many errors, many of them are like `not executable` and `unexpected opcode RefStructGetField` and `unknown type: Ref<S>.` I probably need some help to fix those issues. Let me know what you think. I can still work on the project this weekend.

Best,
Yunqing

titzer:
Ok, I think let's merge this in its current state for now and discuss.

Issue #131, opened 2023-02-17
Improve type system and type inference rules


k-sareen:
As discussed here: https://github.com/titzer/virgil/pull/120#issuecomment-1432581029, the type system and inference rules should be updated to allow less verbose pattern matching for developer UX.

titzer:
Can you write a `seman` testcase that should pass and add to `test/fail`?

k-sareen:
I've been reading the type checking code in `vst/Verifier.v3` and I don't understand why the `ve.expr` has to be `null`, i.e. why the match case has to be the unqualified type (see [these lines](https://github.com/titzer/virgil/blob/e8079d6356f28115954ab5425334d4e6c848a8d1/aeneas/src/vst/Verifier.v3#L1158-L1181)).

k-sareen:
For the error generated by `test/fail/adt_type_infer03.v3` (in PR), I think the [`TypeSystem.isPromotable()`](https://github.com/titzer/virgil/blob/e8079d6356f28115954ab5425334d4e6c848a8d1/aeneas/src/v3/TypeSystem.v3#L75-L94) should return `true` because, well `A<u4>.C` and `A<u8>.C` are the exact same type. Unless is there some internal representation issue? Even then it should just be equal, no?

titzer:
This is a subtle thing that needs to be explained better. Integer types like `u4` and `u8` are not related by *subtyping* but by *promotion*. That is because they could be represented differently. E.g. on a 32-bit machine a `u32` will be represented by one word and `u64` two words. Subtyping is only for values whose representation doesn't change when assigning a value from a subtype to a supertype location (i.e. the assignment is always pass-through, nop).

PR #134, opened 2023-02-18
Add failing tests for ADT type inference

k-sareen:


k-sareen:
As discussed in #131. Some explanations for my reasoning these are type inference bugs:

1. I would argue that `adt_type_infer01.v3` is a type inference bug instead of a syntax bug as it should easily be able to infer that `a` is of type `A<u4>` so it should not be complaining about `MatchError: match on type A<u4> requires unqualified case`.
2. `adt_type_infer02.v3` complains about `TypeError: type "A" expects 1 type arguments and found 0` and `MatchError: key of type A<u4> can never match type A.B<?>` among other things. Again, it should be able to infer that `a` is of type `A<u4>` inside the `match` statement.
3. `adt_type_infer03.v3` is the closest to what we have to do currently to match an ADT with a type parameter, but it can't infer that `A<T>.C` and `A<U>.C` are the same types.

titzer:
I'll tinker with these tests a bit.

titzer:
> As discussed in #131. Some explanations for my reasoning these are type inference bugs:
> 
> 1. I would argue that `adt_type_infer01.v3` is a type inference bug instead of a syntax bug as it should easily be able to infer that `a` is of type `A<u4>` so it should not be complaining about `MatchError: match on type A<u4> requires unqualified case`. This would also allow deconstructing the case expression like with other ADTs.

The rule is that matching on an ADT either deconstructs, in which case you don't qualify the cases, or it is a type-case, in which you need to write full types. It doesn't currently do unification on the target type of a type-case, maybe it should.

> 2. `adt_type_infer02.v3` complains about `TypeError: type "A" expects 1 type arguments and found 0` and `MatchError: key of type A<u4> can never match type A.B<?>` among other things. Again, it should be able to infer that `a` is of type `A<u4>` inside the `match` statement.

This is the unification thing above, it can't figure out the type argument. This test had other bugs; the C case cannot be arbitrarily re-used across different instantiations of A<T>.

> 3. `adt_type_infer03.v3` is the closest to what we have to do currently to match an ADT with a type parameter, but it can't infer that `A<T>.C` and `A<U>.C` are the same types.

They aren't the same types. `A<T>.C` and `A<U>.C` might be represented completely differently, even if the `C` case doesn't mention the type parameter.

> 4. The `adt_type_infer04.v3` test is just a combination of `adt_type_infer01.v3` and `adt_type_infer02.v3`.
> 
> `adt_type_infer0{2,3,4}` use the return value from `g()` in the way we have to do currently. Just added it for sanity that we are using the return value from `g()` and as a test that this kind of matching should be allowed as well (I also didn't want to generate more errors than required).

k-sareen:
> The rule is that matching on an ADT either deconstructs, in which case you don't qualify the cases, or it is a type-case, in which you need to write full types. It doesn't currently do unification on the target type of a type-case, maybe it should.

I mean it would make it less verbose, so definitely better for developer UX.

> This is the unification thing above, it can't figure out the type argument. This test had other bugs; the C case cannot be arbitrarily re-used across different instantiations of A.

Ah right. Interesting. Variants don't get passed by reference, I forgot.

> They aren't the same types. `A<T>.C` and `A<U>.C` might be represented completely differently, even if the `C` case doesn't mention the type parameter.

That's fair. I checked with Rust and yeah I must have been misremembering because an equivalent program does not compile in Rust as well.

PR #141, opened 2023-03-21
Propagate the liveness to the blocks in each loop

s123unny:


titzer:
This one will also need the `aeneas bump`; the two PRs will have to be sequenced, too.

PR #142, opened 2023-03-22
[x86-64] Global register allocator

s123unny:


aeneas/src/mach/GlobalRegAlloc.v3
```diff
@@ -0,0 +1,603 @@
...
+
+	def clear() {
+		visitedCnt = 0;
+		for (i < allocated.length) allocated[i] = (null, -1);
+		for (i < regState.length) regState[i] = -1;
+		for (i < regUsage.length) regUsage[i] = 0;
+		cursor = 0;
+		pos = 0;
+	}
+	def allocate() {
```
titzer, line 40: Let's put this method first and move `clear()` down. Also, add a sentence to `allocate()` to explain the overall algorithm.

aeneas/src/mach/GlobalRegAlloc.v3
```diff
@@ -0,0 +1,603 @@
...
+	}
+	def computeLiveoutForBlock(block: SsaBlock) {
+		liveness.clearRow(liveoutRow);
+		for (e in block.succs()) {
+			var succ = e.dest;
+			// union the succesor's live-in set into this.
+			liveness.or(liveoutRow, succ.mark);
+		}
+	}
+	def allocateFromColor() {
```
titzer, line 63: Maybe a better name is `allocateAfterColoring`?

aeneas/src/mach/GlobalRegAlloc.v3
```diff
@@ -0,0 +1,603 @@
+// Copyright 2023 Virgil Authors. All rights reserved.
+// See LICENSE for details of Apache 2.0 license.
+
+def verbose = false;
+def colorVerbose = false;
+// Implement global register allocation with graph coloring. Build an interference
+// graph from liveness analysis.
+class GlobalRegAlloc(regSet: MachRegSet, codegen: SsaMachGen) {
+	def regState = Array<int>.new(regSet.physRegs);         // indices into {allocated}
```
titzer, line 9: The `regState` data structure seems very similar to that used in `LocalRegAlloc`. We can maybe factor it out in a separate PR.

aeneas/src/mach/GlobalRegAlloc.v3
```diff
@@ -0,0 +1,603 @@
...
+					if (!regSet.isStack(assignment) || !vreg.isConst()) {
+						updateRegState(vreg);
+					}
+				}
+				_ => ;
+			}
+			if (verbose) dumpState("");
+			moves.emit(codegen, cur.next);
+			movesPrev.emit(codegen, cur);
+			if (opcode == ArchInstrs.ARCH_ENTRY) {
```
titzer, line 151: I had another idea to fix the merging of live-in sets from successors (in the interference-graph building phase): make constants interfere at `ARCH_ENTRY`. That may be cheaper overall.

s123unny: In this case, the constant might not interfere with other variable (not defined at `ARCH_ENTRY`, but defined on the path from entry to using constant).

aeneas/src/mach/GlobalRegAlloc.v3
s123unny: Sure. After merging #143, I will modify `GlobalRegAlloc`.

titzer (approved):
lgtm, let's land this!

PR #143, opened 2023-03-23
Factor regState out from localRegAlloc

s123unny:


aeneas/src/mach/RegState.v3
```diff
@@ -0,0 +1,115 @@
+// Copyright 2023 Virgil Authors. All rights reserved.
+// See LICENSE for details of Apache 2.0 license.
+
+class RegState(regSet: MachRegSet, codegen: SsaMachGen) {
```
titzer, line 4: Can you add a few comments for the top-level class and each field to describe what they do?

titzer (approved):
lgtm

PR #149, opened 2023-04-21
Update GlobalRegAlloc

s123unny:
- Fix finding common dominator when inserting reload for spill
- Add coalesce for parallel moves and moves due to constraint (default: false)
- Insert define at basic block/kill for constants
- Add statistic (default: true)


titzer (approved):
lgtm.

PR #155, opened 2023-05-15
[lib] Add apply method for BitMatrix

s123unny:


titzer (approved):
lgtm, thanks.

I'll add a harness for unit tests for this class because one doesn't exist yet. Eventually we should have some unit tests for this class because it's quite tricky inside.

PR #158, opened 2023-05-16
GlobalRegAlloc - Split vars that live across kills before RegAlloc

s123unny:


titzer (approved):
Mostly rubber-stamping the `GlobalRegAlloc` changes, but `ShadowStackSpiller`, and can be refactored with the `BitMatrixRow` along the lines we discussed offline.

PR #159, opened 2023-05-17
Add global-regalloc compile option using matcher

s123unny:
To enable globalRegAlloc: use `-global-regalloc` instead of `-opt=+GlobalRegAlloc`


titzer (approved):
lgtm!

PR #160, opened 2023-05-18
[x86-64]: Add assembler methods for v128 population count

haoyu-zc:
Also added tests to cover previously implemented instructions.


titzer (approved):
lgtm (rubber-stamped)

PR #165, opened 2023-05-23
Insert non-arch-specific instr used in ShadowStackSpiller

s123unny:


titzer (approved):
lgtm; we can even inline some of the codegen methods in the future.

PR #169, opened 2023-06-01
Fix regalloc bug of missing restore in some cases

s123unny:


aeneas/src/mach/GlobalRegAlloc.v3
```diff
@@ -217,15 +211,15 @@ class GlobalRegAlloc(regSet: MachRegSet, codegen: SsaMachGen) {
 			}
 		}
 	}
-	def allocDefReg(vreg: VReg, constraint: int) -> int {
+	def allocDefReg(vreg: VReg, constraint: int, restore: bool) -> int {
```
titzer, line 214: Should we name the new param "genSave"? Because AFAICT that is the behavior that it controls here.

aeneas/src/mach/ShadowStackSpiller.v3
```diff
@@ -295,29 +295,29 @@ class ShadowStackSpiller() {
...
 	}
 	def pushRestoreUp(info: SsaBlockInfo) {
 		var block = info.block;
 		curRow = BitMatrixRow(matrix, reloadStart + block.mark);
 		var succs = block.succs();
+		var killRow = killsFromSucc(block, succs);
 		var child = info.dom_child;
 		if (child != null) {
-			if (child.dom_sibling == null || hasKill(block) || succsHasKill(succs)) {
+			if (child.dom_sibling == null || killRow.count() > 0) {
```
titzer, line 317: We can make this a bit more efficient if instead of `count()` we have `isZero()` for a row.

titzer (commented):
lgtm with some suggestions

titzer (approved):
lgtm

PR #170, opened 2023-06-01
[X86-64]: Add assembler methods for v128 extract_lane and insert_lane

haoyu-zc:
I tried to add two test patterns: `do_s_r_b` and `do_r_s_b" in X86_64AssemblerTestGen.v3. However, when I invoked them with newly implemented instructions `pinsrb_s_r_i` and `pextrb_r_s_i`, the tests failed:

- `pinsrb_s_r_i`: error: invalid combination of opcode and operands
    - Insert a byte integer value from r32/m8 into xmm1 at the destination element in xmm1 specified by imm8.
- `pextrb_r_s_i`: Comparing...failed
    - Extract a byte integer value from xmm2 at the source byte offset specified by imm8 into reg or m8. The upper bits of r32 or r64 are zeroed.

For both of the two instructions I'm sure I wrote the bytes correctly. But I have no clues what would lead to the failures. Could you give me some help here?

### DIFF
--- a/lib/asm/x86-64/X86_64Assembler.v3
+++ b/lib/asm/x86-64/X86_64Assembler.v3
@@ -1835,7 +1835,7 @@ class X86_64Assembler(w: DataWriter, OP_REX: byte) {
 		emitb(0x66);
 		emit_rex_bb_r_r(a, b, NO_REX, 0x0F, 0x6C);
 	}
-	def pshufd_s_s_imm(a: X86_64Xmmr, b: X86_64Xmmr, imm: u8) -> this {
+	def pshufd_s_s_i(a: X86_64Xmmr, b: X86_64Xmmr, imm: u8) -> this {
 		emitb(0x66);
 		emit_rex_bb_r_r(a, b, NO_REX, 0x0F, 0x70);
 		emitb(imm);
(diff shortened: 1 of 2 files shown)


titzer:
If those are lane immediates, then `nasm` is probably checking that they are in range of the lane indices. So you should change the assembler test generator to have a set of valid indices.

haoyu-zc:
Thank you! It turns out I need to set `regSize` to specify r32 registers

haoyu-zc:
I'm stuck with another issue now. I created a new test pattern `do_r_s_b` for lane extractions but the tests failed. One of the assembler methods as an example is:

```C
def pextrq_r_s_i(a: X86_64Gpr, b: X86_64Xmmr, imm: u8) -> this {
	emitb(0x66);
	emit_rex_bbb_r_r(a, b, REX_W, 0x0F, 0x3A, 0x16);
	emitb(imm);
}
```

However, the generated asm code is off at the bytes specifying registers in some lines:
image

I doubt the method `emit_rex_bbb_r_r` is not compatible with patterns like `r_s_b". Or could you give me some suggestions to fix this? Thank you!

(The diff file is attached)

[test.list.diff.txt](https://github.com/titzer/virgil/files/11639516/test.list.diff.txt)

titzer:
Hmm, I don't have any more insights other than that instruction is apparently not encoded the way the utility method works. You'll have to double-check how it is encoded in the manual.

haoyu-zc:
- Added new utility methods `emit_rex_bbb_r_r2` and `emit_rex_bbb_m_r` to correctly handle v128 lane extractions

PR #176, opened 2023-06-13
[wasi]: add openat systemcall in LinuxConst.v3

pco2699:
`openat` and `openat2` system calls are required for wasi implementation in https://github.com/titzer/wizard-engine/, so I added some constants in `LinuxConst.v3`


pco2699:
Based on the discussion, we decided to implement `openat` in wizard itself, so I'll just close this pull request.

PR #187, opened 2023-07-25
Fix V3.renderBottom

s123unny:
To reproduce the bug: `v3c -run -ssa-int -trace test/core/array_tuple21.v3`


titzer (changes requested):
lgtm; can you bump the version number too?

PR #195, opened 2023-08-16
Add line number information with Dwarf in elf executable

s123unny:


aeneas/src/os/Linux.v3
```diff
@@ -142,32 +143,40 @@ class LinuxTarget extends Target {
...
-				m.source.renderLong(buf);
-				buf.out(w.putkv);
-				w.putb(0);
+			if (CLOptions.SYMBOLS.val) {
+				sections.addFuncSymbol(w, ".entry", mach.entryStub);
+				if (mach.allocStub != null) sections.addFuncSymbol(w, ".alloc", mach.allocStub);
+				if (mach.signalStub != null) sections.addFuncSymbol(w, ".signal", mach.signalStub);
+
+				var buf = StringBuilder.new();
+				for (i < prog.ir.methods.length) {
```
titzer, line 152: Now that this is nested another level, it's probably a good idea for us to factor out a method `addMethodSymbol()`.

titzer (approved):
lgtm with minor suggestion

aeneas/src/os/Linux.v3
```diff
@@ -285,6 +263,31 @@ class ElfSections(elf: ElfHeader) {
...
 		str.sh_size = w.pos - str_start;
 	}
+	// Add symbol entry and write method name to .strtab section
+	def addMethodSymbol(w: MachDataWriter, mach: MachProgram, prog: Program) {
+		addFuncSymbol(w, ".entry", mach.entryStub);
+		if (mach.allocStub != null) addFuncSymbol(w, ".alloc", mach.allocStub);
+		if (mach.signalStub != null) addFuncSymbol(w, ".signal", mach.signalStub);
+
+		var buf = StringBuilder.new();
+		for (i < prog.ir.methods.length) {
```
titzer, line 273: Oh, I meant just to pull out the body of the loop that works on a single method.

s123unny: What about the stringBuilder? or use `addFuncSymbol()` as well?

titzer: Yeah, you can pass in the StringBuilder as a temporary. I didn't quite understand the `addFuncSymbol()` comment. If you go back to before extracting this method and only extract the loop body, I think that solves that.

titzer (approved):
lgtm, thanks!

PR #196, opened 2023-08-20
Change SsaInterpreter frame to link-list

s123unny:


aeneas/src/ssa/SsaInterpreter.v3
```diff
@@ -2,13 +2,11 @@
...
-def INITIAL_FRAMES = 16;
 
 // Implements an interpreter capable of executing all SSA instructions, calls, etc
 // for initialization and testing.
 class SsaInterpreter(prog: Program, genSsa: (IrSpec, int) -> SsaGraph) {
 	var values = Array<Val>.new(INITIAL_VALUES);
-	var stack = Array<SsaIntFrame>.new(INITIAL_FRAMES);
 	var frame: SsaIntFrame;
 	def args = SsaIntArguments.new(prog);
 	var sp: int;
```
titzer, line 12: Is it possible to remove `sp` as well?

titzer (changes requested):
lgtm with minor suggestion

aeneas/src/ssa/SsaInterpreter.v3
s123unny: `traceInstr()` uses `sp` to decide indent

titzer: What about just walking the frames? It's already linear in the number of frames, since it counts from `0` to `sp`.

PR #198, opened 2023-09-04
[WIP] Refactoring RefLayout

btwj:
The RefLayout{Get/Set}Field opcodes (and their repeated counterparts) would read/write one byte at a time.
This PR introduces two new Opcodes that are created during normalization:
- `ByteArrayGetField[offset: int]<T>(arr: Array<byte>, i_offset: int) -> T`
- `ByteArraySetField[offset: int]<T>(arr: Array<byte>, i_offset: int, value: T) -> void`

All RefLayout reads/writes (including those for repeated fields) are normalized into this form.

When lowering to machine code, we decompose these loads/stores into power-of-two loads/stores (e.g. a `u56`, with 7 bytes, will be decomposed into loads of 4+2+1 bytes).

Todo:
- Since there is no intermediate lowering for JVM code generation, add a normalizer flag to perform byte-by-byte loads/stores (same as the old behavior) specially for the JVM target.


aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -390,60 +390,47 @@ class SsaRaNormalizer extends SsaRebuilder {
...
-						var wt = Int.getType(false, x.byteSize() * 8);
-						result = curBlock.opIntViewI0(wt, x, result);
-						return map1(app, result);
-					}
+					x: IntType => result = curBlock.opByteArrayGetField(x, offset, facts, array, i_offset);
+					x: FloatType => result = curBlock.opByteArrayGetField(x, offset, facts, array, i_offset);
 					x: EnumType => {
 						var it = IntType.!(x.enumDecl.tagType);
-						var result = genArrayByteGetMultiple(app, it, array, i_offset, offset);
+						var wt = Int.getType(false, it.byteSize() * 8);
```
titzer, line 401: Why not just use the tag type `it` here?

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -390,60 +390,47 @@ class SsaRaNormalizer extends SsaRebuilder {
...
 					x: EnumType => {
 						var it = IntType.!(x.enumDecl.tagType);
-						genArrayByteSetMultiple(app, it, array, i_offset, offset, val);
-					}
-					x: FloatType => {
-						var it = Int.getType(false, x.total_width);
-						var width = x.byteSize();
-						val = curBlock.opIntView(x, it, val);
-						genArrayByteSetMultiple(app, it, array, i_offset, offset, val);
+						var wt = Int.getType(false, it.byteSize() * 8);
```
titzer, line 429: Same here. Why not a write of the tag type?

aeneas/src/mach/MachLowering.v3
```diff
@@ -1862,6 +1891,38 @@ class MachLowering(mach: MachProgram, compiler: Compiler, config: MachLoweringCo
...
+	def genNormTypedLoadsByteRegion(source: Source, machType: Type, base: SsaInstr, offset: int) -> Array<SsaInstr> {
+		var tn: TypeNorm;
+		match (machType) {
+			x: IntType => tn = mach.intNorm.makeType(x);
+			_ => return [ptrLoad(machType, base, offset)];
+		}
+
+		var loads = Array<SsaInstr>.new(tn.size);
+
+		for (i < loads.length) {
```
titzer, line 1903: As discussed, I think you only need to worry about handling partial reads for the last `tn.sub[i]`. You can over-read for a field that is not too close to the end of a struct. We can do that optimization later and control it with a `Fact` set on the instruction.

rt/jvm/Makefile
```diff
@@ -1,5 +1,5 @@
 bin/V3S_System.class: src/V3S_System.java
-	javac -source 1.5 -target 1.5 -d bin/ src/V3S_System.java
```
titzer, line 2: Unless there is a good reason for this, I would prefer to just keep it 1.5.

aeneas/src/core/Eval.v3
```diff
@@ -844,13 +844,23 @@ def evalOp(op: Operator, args: Arguments) -> Result {
...
 			var ref = args.ref(0);
 			var index = args.i(1);
 			if (u32.view(index) >= u32.view(max)) return args.throw(V3Exception.BoundsCheck, null);
-			return doRefLayoutSetField(args, ref, offset + scale * index, args.getArg(2));
+			return doRefLayoutSetField(args, args.getTypeArg(1), ref, offset + scale * index, args.getArg(2));
+		}
+		ByteArrayGetField(offset) => {
+			var array = args.r(0);
+			var i_offset = args.i(1);
+			return doRefLayoutGetField(args, args.getTypeArg(0), ByteArrayOffset.new(array, offset), i_offset);
```
titzer, line 858: Can you refactor the underlying doRefLayoutGetField to take the array and offset individually to avoid creating an intermediary `ByteArrayOffset` object?

aeneas/src/mach/MachLowering.v3
```diff
@@ -1862,6 +1891,38 @@ class MachLowering(mach: MachProgram, compiler: Compiler, config: MachLoweringCo
...
 	}
+	def genNormTypedLoadsByteRegion(source: Source, machType: Type, base: SsaInstr, offset: int) -> Array<SsaInstr> {
+		var tn: TypeNorm;
+		match (machType) {
+			x: IntType => tn = mach.intNorm.makeType(x);
+			_ => return [ptrLoad(machType, base, offset)];
+		}
+
+		var loads = Array<SsaInstr>.new(tn.size);
+		for (i < tn.size) {
```
titzer, line 1902: What about factoring out the first `tn.size - 1` into its own loop? (and also, AFAICT, those don't need to be `IntType`, as `mach.sizeOf()` should handle the other cases)

rt/jvm/src/V3S_System.java
```diff
@@ -429,4 +429,151 @@ public static boolean feq(float a, float b) {
     public static boolean deq(double a, double b) {
 	return Double.doubleToRawLongBits(a) == Double.doubleToRawLongBits(b);
     }
+
+    public static byte readBytes1(byte[] bytes, int offset, int staticOffset) {
+        return bytes[offset + staticOffset];
+    }
+
+    public static short readBytes2(byte[] bytes, int offset, int staticOffset) {
+        int byte0 = ((int)bytes[offset + staticOffset] & 0xFF);
```
titzer, line 438: What about factoring out `offset + staticOffset` into a (short) temp? That shortens the source a bit and will also save some bytecode space.

titzer (approved):
lgtm other than small comments

aeneas/src/core/Eval.v3
btwj: The underlying `doRefLayoutGetField` is taking an offset since the (array, offset) pair is being passed around as a single value when evaluating without normalization. There are calls to methods on `ByteArrayOffset` in `doRefLayoutGetField`, so we'll have to lift those methods out of the class to avoid creating an object. Should we still do it?

titzer: Ok, yeah, I see that now. It's OK to keep as is and drop in an `XXX`.

titzer (approved):

PR #203, opened 2023-09-26
Fix Wasm layout bug

btwj:


titzer (approved):
lgtm; can you bump the version number as well?

titzer (approved):

PR #209, opened 2023-11-06
Variant Unboxing

btwj:
This PR enables (with the `-unbox-variants` flag) the ability to unbox ADTs, both at the parent level and at a per-case level.
For now, these ADTs are normalized by greedily packing fields into the fewest scalars possible, and then attaching an extra scalar representing the tag.

Todo:
- Add more tests for large ADTs
- Add unit tests for VariantSolver


aeneas/src/ir/VariantSolver.v3
```diff
@@ -0,0 +1,330 @@
+// Norm for variants (both superclass and case class)
+class VariantNorm extends TypeNorm {
```
titzer, line 2: There are a lot of different classes now, and it looks like `VariantNorm` and `VariantParentNorm` are basically just markers (i.e. common superclasses) for the several subclasses below. I generally try to avoid that. Can we reduce the number of classes and maybe use a field to distinguish between an unboxed data norm and an unboxed case norm? (i.e. the only different right now is that the case norm has a tag).

btwj:
@titzer I've gone ahead and added `#unboxed` to `type Operand` here, it bootstraps now!

Also fixed a bug in the existing compiler that I discovered while writing extra tests: equality for boxed variants in tuples wasn't normalized correctly, e.g. if `U` was a boxed variant, the instruction `OverloadedEq<(U, U)>` ended up becoming normalized wrongly as `VariantEq<U>` when it should've become `CallVariantSelector<U>[==]`. See `test/variants/eq00.v3` for an example that crashes the old compiler.

Separately, the equality code for `U` wasn't deemed reachable from `OverloadedEq<(U, U)>` because `getEquality` doesn't recurse into the nested types if the type was a tuple, I've added that as well.

titzer:
Superseded.

PR #210, opened 2023-11-28
Check for test failure for each internal target

### LINKED ISSUE #208: CI Scripts don't report failure properly
See: https://github.com/titzer/virgil/actions/runs/6606436467/job/17942642921

A `test/range/` test failed but the overall CI run did not fail because of this.

k-sareen:
Closes #208.

### DIFF
--- a/test/common.bash
+++ b/test/common.bash
@@ -353,16 +353,16 @@ function execute_target_tests() {
 function execute_tests() {
     for target in $TEST_TARGETS; do
 	if [ "$target" = "v3i" ]; then
-	    execute_v3i_tests "v3i" ""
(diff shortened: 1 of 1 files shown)


k-sareen:
The issue was that the `v3i` target was failing but we were returning the error code for the `v3i-ra-ma` internal target. We should do a failure check for each of the commands to ensure we report failures correctly and fail as soon as possible.

k-sareen:
If you checkout to the offending commit (https://github.com/titzer/virgil/commit/4868eae743511698439e179983521350dd5a9b90) and then cherry-pick this fix, you'll find it fails correctly as expected.

titzer (approved):
lgtm, thanks!

PR #214, opened 2024-01-17
Adjust RepHint Parsing

btwj:


aeneas/src/vst/Parser.v3
```diff
@@ -185,18 +185,28 @@ component Parser {
...
 		file.classes.put(decl);
 		// lift case classes to top-level of file
 		for (cdecl in decl.cases) {
 			if (cdecl.decl != decl) file.classes.put(cdecl.decl);
 		}
 		return true;
 	}
-	def parseRepHint(p: ParserState) -> VstRepHint {
-		var list: List<VstIdent<void>>;
+	def parsePacking(p: ParserState) -> string {
```
titzer, line 198: This eats far too much as a token, e.g. all the normal lexical delimiters like `(){}`, and keywords and so on. We need to be conservative by default and only parse packings that we intend to support. E.g. I think we want to support packing expressions that reference a declared packing (with arguments as variables or constants), but not much else.

aeneas/src/vst/Parser.v3
```diff
@@ -185,18 +185,44 @@ component Parser {
...
+	}
+	def parsePacking(p: ParserState) -> VstPacking {
+		if (p.curByte == '0') {
+			p.eat1();
+			p.req1('b');
+			var s = p.curPos, q = p.curPos;
+			var chars = Vector<u8>.new();
+			while (q < p.input.length) {
+				var c = p.input[q];
+				if (c == '0' || c == '1' || Char.isIdentStart(c)) chars.put(c);
```
titzer, line 209: Should we allow '?' here too?

aeneas/src/vst/Parser.v3
```diff
@@ -185,18 +185,44 @@ component Parser {
...
+				else if (c != '_') break;
+				q++;
+			}
+			p.advance(q - s);
+			return VstPacking.Bits(chars.array);
+		} else if (Char.isIdentStart(p.curByte)) {
+			var packingName = parseIdentVoid(p);
+			var fields = parseList(0, p, '(', COMMA, ')', parseIdentVoid);
+			return VstPacking.Params(packingName.name, fields.map(asToken));
```
titzer, line 218: You can use `VstIdent<void>.name` as a getter method here.

aeneas/src/vst/Parser.v3
```diff
@@ -207,10 +233,18 @@ component Parser {
 				q++;
 			}
 			var id = extractIdent<void>(p, q);
-			if (id.kwClass != 0) kwError(p, id.name);
-			list = List.new(id, list);
+			var name = id.name.image;
+			var result: VstRepHint;
+			if (Strings.equal(name, "boxed")) result = VstRepHint.Boxed;
+			else if (Strings.equal(name, "unboxed")) result = VstRepHint.Unboxed;
+			else if (Strings.equal(name, "packing")) {
```
titzer, line 240: Packing is a keyword now; let's introduced `KC_PACKING` to allow it (but not other keywords) here.

aeneas/src/vst/Vst.v3
```diff
@@ -375,9 +375,16 @@ class SuperClause {
...
+type VstRepHint {
+	case Boxed;
+	case Unboxed;
+	case Packing(p: VstPacking);
+	case Other(hint: string);
+}
+
+type VstPacking {
+	case Bits(rep: string);
+	case Params(packingName: Token, fields: VstList<Token>);
```
titzer, line 387: What about making the fields a `VstList<Expr>`? That way we can accommodate `VarExpr` and `Literal`.

titzer:
Can you bump the version number and I'll merge this?

titzer (approved):

PR #218, opened 2024-02-09
Shift DataNorm representation to VariantNorm

btwj:
This PR shifts the variant logic into `VariantSolver.v3`, and migrates away from the use of `DataNorm` to `VariantNorm`. `VariantNorm`s are able to represent both enum-like variants (no fields) and single-case, flattened variants.
This does not add any additional functionality.

### DIFF
--- a/aeneas/src/ir/Normalization.v3
+++ b/aeneas/src/ir/Normalization.v3
@@ -33,6 +33,9 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 	var specializer: Specializer;
 	var virtuals: List<RaVirtual>;
 	var ovfAlloc: OverflowFieldAllocator;
+	var vs: VariantSolver;
+
+	new() { vs = VariantSolver.new(config, this, /* CLOptions.PRINT_RA.get() */ true); }
 
 	def normalize() {
 		if (CLOptions.PRINT_DEAD_CODE.get()) DeadCodeAnalyzer.new(ra).report();
@@ -83,7 +86,7 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 				ra.queue.add(normClassRecord, (rc, oldRecord, newRecord.values));
 			}
 			ra.prog.setComponentRecord(comp, newRecord);
-		} else if (rc.dataNorm == null) {
+		} else if (rc.variantNorm == null) {
 			// create and map new records to be normalized
 			for (l = rc.instances; l != null; l = l.tail) {
 				var oldRecord = l.head, newRecord = ra.prog.newRecord(tn.newType, rc.normFields.length);
@@ -105,9 +108,9 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 				var rm = l.head, m = rm.orig;
 				if (rm.norm != null) continue; // already done
 				var ftype = if(rm.spec == null, m.sig.funcType(), rm.spec.getMethodType());
-				if (rc.raFacts.RC_UNBOXED) {
+				if (rc.variantNorm != null) {
 					// move flattened data type receiver to function sig
-					ftype = Function.prependParamTypes(rc.dataNorm.sub, ftype);
+					ftype = Function.prependParamTypes(rc.variantNorm.sub, ftype);
 				}
 				rm.funcNorm = FuncNorm.!(norm(ftype));
 				var typeParams = if(rm.spec != null, rm.spec.getTypes().methodTypeArgs);
@@ -205,7 +208,7 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 			V3Kind.VARIANT => {
 				// try flattening variants and data types
 				var rc = ra.getClass(t);
-				if (rc != null) tn = normVariant(t, rc);
+				if (rc != null) tn = vs.normVariant(t, rc);
 				else tn = TypeNorm.new(t, t, null);
 			}
 			V3Kind.ENUM => {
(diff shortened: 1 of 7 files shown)


aeneas/src/ir/Normalization.v3
```diff
@@ -421,18 +334,35 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
...
 		}
-		var rfs = rc.fields;
-		for (i < rfs.length) {
-			var rf = rfs[i];
-			if (rf != null && rf.normIndex >= 0) {
-				var v = oldRecord.values[i];
-				if (rf.fieldType == null) array[rf.normIndex] = normSimpleVal(null, v);
-				else normValIntoArray(v, rf.typeNorm, array, rf.normIndex);
+
+		if (!vn.isEnum()) {
```
titzer, line 351: What about returning early if it's an enum? (Thus the following code is indented less.

aeneas/src/ir/Normalization.v3
```diff
@@ -421,18 +334,35 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
...
+			var rfs = vrc.fields;
+			for (i < vrc.fields.length) {
+				var rf = vrc.fields[i];
+				if (rf != null && rf.normIndex >= 0) {
+					var v = oldRecord.values[i];
+					var indexes = vn.fields[i].indexes;
+					var fieldArray = Array<Val>.new(indexes.length);
```
titzer, line 358: `fieldArray` is temporary here, right? Maybe add an XXX to think about how to reuse that temp.

aeneas/src/ir/Normalization.v3
```diff
@@ -461,7 +391,7 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 	def normSimpleVal(tn: TypeNorm, v: Val) -> Val {
 		match (v) {
 			x: Record => {
-				if (DataNorm.?(tn)) return normDataVal(DataNorm.!(tn), x)[0];
+				if (tn != null && VariantNorm.?(tn) && VariantNorm.!(tn).isEnum()) return normVariantVal(VariantNorm.!(tn), x)[0];
```
titzer, line 394: `VariantNorm.?(tn)` will be false for `null`.

aeneas/src/ir/Normalization.v3
```diff
@@ -620,7 +550,7 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 	}
 	def createIrClass(rc: RaClass) {
 		var sc = if(rc.parent != null, rc.parent.normClass);
-		var normFields = if(rc.raFacts.RC_UNBOXED, NO_FIELDS, rc.normFields);
+		var normFields = if(rc.variantNorm != null, NO_FIELDS, rc.normFields);
```
titzer, line 553: Checking `.variantNorm != null` seems common. What about a method `RaClass.isUnboxed()` ?

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -1209,15 +1218,13 @@ class SsaRaNormalizer extends SsaRebuilder {
 			if (!isVariant) addNullCheck(oldApp, ninputs[0]);
 			return map0(oldApp);
 		}
-		if (raField.raFacts.RC_UNBOXED) {
-			// field of flattened data type
-			normType(raField.receiver); // XXX: normType() side-effect of flattening
-			var rc = norm.ra.getClass(raField.receiver);
+		normType(raField.receiver); // XXX: normType() side-effect of flattening
```
titzer, line 1221: Can you make use of the return value from `normType`?

aeneas/src/ir/VariantSolver.v3
```diff
@@ -0,0 +1,171 @@
...
+		buf.puts(")");
+		if (tn != null) buf.put1(": [%q]", tn.newType.render);
+		return buf;
+	}
+}
+
+def ON_STACK = -1;
+
+class VariantSolver(nc: NormalizerConfig, rn: ReachabilityNormalizer, verbose: bool) {
+	def normVariant(t: Type, rc: RaClass) -> TypeNorm {
```
titzer, line 68: Can you documents the methods of this class? It's the important interface between the `Normalizer` and the solver.

aeneas/src/main/CLOptions.v3
```diff
@@ -149,8 +149,12 @@ component CLOptions {
 		"Normalize ranges using RangeStart's to support off-heap ranges.");
 	def SET_EXEC		= compileOpt.newBoolOption("set-exec", true,
 		"Automatically set execute permission for compiled binaries.");
-	def USE_GLOBALREGALLOC = compileOpt.newMatcherOption("global-regalloc",
+	def USE_GLOBALREGALLOC	= compileOpt.newMatcherOption("global-regalloc",
 		"Enable global register allocator.");
+	def UNBOX_VARIANTS	= compileOpt.newBoolOption("unbox-variants", false,
+		"Enable variant unboxing features.");
+	def UNBOX_ALL		= compileOpt.newBoolOption("unbox-all", false,
```
titzer, line 156: What about `-unbox-variants=<filter>` that accepts a glob, with the default empty pattern being "all"?

Also, about what `-unbox-variant-cases=<list>` that takes a list of integers and marks all corresponding cases as `#unboxed`? That would give a fuzzer or a manual tester a lot of control to try out many combinations.

titzer (commented):
This is looking good, mostly small comments.

aeneas/src/ir/SsaNormalizer.v3
btwj: I don't think using the return value here will make much of a difference - and I'll have to cast it to a `VariantNorm` to use it in the if body below anyway (as opposed to just using `rc.variantNorm`)

aeneas/src/main/CLOptions.v3
btwj: I'll put the implementation of the `-unbox-variant-cases=<list>` in the PR with case-specific unboxing, but I've changed the appropriate variable here.

titzer (approved):
lgtm

PR #219, opened 2024-02-13
Packing Declaration Parsing

btwj:
Introduces `VstPackingDecl` and `VstPackingExpr`, allowing for nested packing expressions and packing literals.

### DIFF
--- a/aeneas/src/vst/Parser.v3
+++ b/aeneas/src/vst/Parser.v3
@@ -196,23 +196,38 @@ component Parser {
 		}
 		return true;
 	}
-	def parsePacking(p: ParserState) -> VstPacking {
-		if (p.curByte == '0') {
-			p.eat1();
-			if (p.curByte == 'b' || p.curByte == 'B') p.eat1();
-			else p.error("expected bit pattern");
+	def parsePackingParam(p: ParserState) -> (Token, u8) {
+		var param = parseIdentVoid(p).name;
+		p.req1(':');
 
-			var s = p.curPos, q = p.curPos;
-			var chars = Vector<u8>.new();
-			while (q < p.input.length) {
-				var c = p.input[q];
-				if (c == '0' || c == '1' || c == '?' || Char.isIdentStart(c)) chars.put(c);
-				else if (c != '_') break;
-				q++;
-			}
-			p.advance(q - s);
-			return VstPacking.Bits(chars.array);
-		} else if (Char.isIdentStart(p.curByte)) {
+		var sizeLiteral = parseIntLiteral(p);
+		var size = u8.!(V3.unboxI32(sizeLiteral.val));
+		return (param, size);
+	}
+	def parsePackingBits(p: ParserState) -> string {
+		p.req1('0');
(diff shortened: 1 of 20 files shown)


aeneas/src/vst/Parser.v3
```diff
@@ -196,30 +198,64 @@ component Parser {
...
+		var size = u8.!(V3.unboxI32(sizeLiteral.val));
+		return (param, size);
+	}
+	def parsePackingSize(p: ParserState) -> u8 {
+		var sizeLiteral = parseIntLiteral(p);
+		if (sizeLiteral == null) {
+			p.error("size expected");
+			return 0;
+		}
+		else return u8.!(V3.unboxI32(sizeLiteral.val));
```
titzer, line 215: I think we want to keep this as a literal, so that it can be checked in the Verifier that it's a reasonable size.

aeneas/src/vst/Vst.v3
```diff
@@ -378,13 +380,24 @@ class SuperClause {
...
 	case Boxed;
 	case Unboxed;
-	case Packing(p: VstPacking);
+	case Packing(p: VstPackingExpr);
 	case Other(hint: string);
 }
 
-type VstPacking {
+type VstPackingExpr {
 	case Bits(rep: string);
```
titzer, line 388: Should we make a variant here for the bits and store an array of them? E.g.

```
type PackingBit {
  case Literal(val: u1);
  case Wildcard;
  case Char(ch: byte);
}
```

aeneas/src/vst/Parser.v3
```diff
@@ -196,30 +198,64 @@ component Parser {
...
+	}
+	def parsePackingExpr(p: ParserState) -> VstPackingExpr {
+		var s = p.curPos, q = p.curPos;
+		var isBitPattern = false;
+		if (p.peek(0) == '0' && p.peek(1) == 'b') {
```
titzer, line 220: Probably want to allow uppercase 'B' as well here.

test/variants/parser/packing08.v3
```diff
@@ -0,0 +1,2 @@
+//@parse
```
titzer, line 1: Can you add some negative parsing tests as well? Also, some simpler cases with only one field, some parts are missing, etc?

aeneas/src/vst/VstPrinter.v3
```diff
@@ -104,24 +104,51 @@ class Printer(printer: VstPrinter) {
...
 			Other(s) => Terminal.put1("#%s", s);
 		}
 	}
 	def printToken(t: Token) {
 		Terminal.put1("%s", t.image);
 	}
-	def printPacking(p: VstPacking) {
+	def printPackingBits(bits: Array<PackingBit>) {
+		Terminal.put("0b");
+		for (bit in bits) {
```
titzer, line 117: Maybe add a `_` every four or eight bits?

titzer (approved):
Logic looks good, just needs more tests and a bump of the version number.

titzer (approved):

PR #224, opened 2024-03-13
[WIP] Packing Solver + Unit Tests

btwj:


btwj:
@titzer Missing a couple validity checks in `PackingSolverTest`, but any comments so far?

aeneas/src/ir/PackingSolver.v3
```diff
@@ -0,0 +1,332 @@
...
+		for (j = i.s; j < i.e; j++) bits[j] = Bit.Unusable;
+	}
+	def unassignInterval(i: Interval) {
+		for (j = i.s; j < i.e; j++) bits[j] = Bit.Unassigned;
+	}
+	def copy() -> PackingState { return PackingState.new(Arrays.dup(bits)); }
+}
+
+type PackingField {
+	case Nonref(size: int);
```
titzer, line 125: This is the field size in bits, correct? I think using a `byte` here is reasonable; it will allow up to 255 bits.

aeneas/src/ir/PackingSolver.v3
```diff
@@ -0,0 +1,332 @@
...
+	case Ref;
+}
+
+type FieldLoc(caseIdx: int, fieldIdx: int) {
+	def hash() -> int { return caseIdx * 31 + fieldIdx; }
+
+	def equal(other: FieldLoc) -> bool { return caseIdx == other.caseIdx && fieldIdx == other.fieldIdx; }
```
titzer, line 132: Because this is a datatype, the builtin `==` will do a structural check, so this method isn't needed.

aeneas/src/ir/PackingSolver.v3
```diff
@@ -0,0 +1,332 @@
...
+	}
+	def copy() -> PackingState { return PackingState.new(Arrays.dup(bits)); }
+}
+
+type PackingField {
+	case Nonref(size: int);
+	case Ref;
+}
+
+type FieldLoc(caseIdx: int, fieldIdx: int) {
```
titzer, line 129: Maybe call this a `CaseField`?

aeneas/src/ir/PackingSolver.v3
```diff
@@ -0,0 +1,332 @@
...
+	def hash() -> int { return caseIdx * 31 + fieldIdx; }
+
+	def equal(other: FieldLoc) -> bool { return caseIdx == other.caseIdx && fieldIdx == other.fieldIdx; }
+}
+
+class PackingSolver(size: u8, gcPatterns: GcPatterns) {
```
titzer, line 135: Maybe `refPatterns` instead of `gcPatterns`?

aeneas/src/ir/PackingSolver.v3
```diff
@@ -0,0 +1,332 @@
...
+		this.cases = cases;
+		
+		var solution = solve(0, initialState);
+		if (solution == null) return null;
+
+		return assignments;
+	}
+}
+
+type GcPatterns(scanned: PackingState, refInterval: Interval, ignored: PackingState);
```
titzer, line 320: What about the terms `ptrref` and `nonptrref`? Both are reference words (i.e. marked as a reference in stackmaps and object maps), but `ptrref`s contain addresses that are followed by the GC.

aeneas/src/ir/PackingSolver.v3
```diff
@@ -0,0 +1,332 @@
...
+			}
+			Fixed(x) => match (b) {
+				Unusable => return (a, false);
+				Unassigned => return (a, true);
+				Fixed(y) => return (a, x == y);
+			}
+		}
+	}
+	def mergeBits(a: Array<Bit>, b: Array<Bit>) -> Array<Bit> {
+		if (a.length != b.length) return null;
```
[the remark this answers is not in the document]
btwj: This part is dead code from an old implementation, I'll get rid of it.

aeneas/src/ir/PackingSolver.v3
```diff
@@ -0,0 +1,339 @@
...
+class PackingSolution(patterns: Array<PackingPattern>, assignments: HashMap<CaseField, Interval>, isRef: bool, problem: PackingProblem) {}
+
+type RefPatterns(
+	ptrref: PackingPattern,
+	refInterval: Interval,
+	nonptrref: PackingPattern,
+	nonref: PackingPattern
+);
+
+component MachPatterns {
```
titzer, line 326: This should definitely be part of the `Target`.

aeneas/test/PackingSolverTest.v3
```diff
@@ -0,0 +1,192 @@
+// Copyright 2024 Virgil authors. All rights reserved.
+// See LICENSE for details of Apache 2.0 license.
+
+def T = UnitTests.registerT("packing:", _, PackingSolverTester.new, _);
+def X = [
```
titzer, line 5: I think we could use some tests for `canDistinguish` as well.

aeneas/src/ir/PackingSolver.v3
```diff
@@ -0,0 +1,339 @@
...
+		assignments = HashMap<CaseField, Interval>.new(CaseField.hash, CaseField.==);
+	}
+
+	def canDistinguish(state: Array<PackingPattern>, elements: Array<bool>) -> bool {
+		// TODO: split on bits with the most unassigned
+		var numElements = 0;
+		for (i in elements) if (i) numElements++;
+		if (numElements <= 1) return true;
+
+		for (i < size) {
```
titzer, line 126: I think this will be more efficient if you do a first pass over all the bits and try to find a bit that is fixed in all patterns, in order to avoid duplicating an element when trying an assignment, ranking them otherwise along the way. If you find a bit fixed in all elements, stop and use it right away. If no bit fixed in all patterns is found, then use the best-ranked bit instead. The rank is some function of how many patterns have that unassigned versus fixed 0 or 1. E.g. picking a bit that is fixed in most patterns (but all 0) isn't particularly helpful;there's probably an information-theoretic best choice. But obviously it's a heuristic, so any choice is going to be bad in some situations.

aeneas/src/ir/PackingSolver.v3
```diff
@@ -0,0 +1,339 @@
...
+				}
+			}
+
+			if (canDistinguish(state, zeros) && canDistinguish(state, ones)) return true;
+			for (j < temporaryBits.length) state[temporaryBits[j]].bits[i] = Bit.Unassigned;
+		}
+		return false;
+	}
+
+	def checkDistinguishable(state: Array<PackingPattern>) -> bool {
```
titzer, line 171: This has side-effects, so maybe `trySimpleTaggingHeuristic` and then split off the other heuristic into another method?

aeneas/src/ir/PackingSolver.v3
```diff
@@ -0,0 +1,339 @@
...
+				if (state[j].bits[i] != Bit.Unassigned) {
+					allUnassigned = false;
+					break;
+				}
+			}
+			if (allUnassigned) unassigned.put(i);
+		}
+
+		var result = false;
+		if (1 << byte.!(unassigned.length) >= state.length) {
```
titzer, line 187: I guess we expect that the simple tagging heuristic will normally work. If it does work, then there are bits that represent the language-level tag. Thus the `VariantGetTag` will want to know which ones they are. Maybe we need to store them in the solution somehow?

aeneas/src/ir/PackingSolver.v3
btwj: I've kept these values as a component, and set the value of `refPattern` in the constructor in `Linux.v3`. Wdyt?

titzer (approved):
Ok, looks good enough to land. I'll take a stab at the verifier implementation for packing declarations.

PR #225, opened 2024-03-20
[packing] Modifications to Packing Parsing + Verifying

btwj:
Makes the following changes to packing parsing:
- Packings in declarations no longer support tuples, since declarations should not contain unsolved fields. Tuples are now `#solve` expressions, and concatenations are `#concat` expressions.
- Separating scalars is done with `#packing (x, y)` which is _not_ represented as a tuple expression, but a list in the packing `repHint`.
- Verify packing applications.


aeneas/src/vst/Parser.v3
```diff
@@ -231,28 +238,50 @@ component Parser {
...
+				result = VstPackingExpr.App(varExpr, parseList(0, p, '(', COMMA, ')', parsePackingExpr(_, inDecl, true)));
+			} else {
+				result = VstPackingExpr.Field(varExpr);
+			}
+		} else {
+			result = VstPackingExpr.Literal(parseNumber(p));
+		}
+
+		if (p.curByte == '<') {
+			p.reqN("<<|");
```
titzer, line 268: You'll want to keep the token for this for a location to report an error.

aeneas/src/vst/Verifier.v3
```diff
@@ -93,6 +93,12 @@ class Verifier(compiler: Compiler, prog: Program) {
 		for (i < file.packings.length) {
 			var packingDecl = file.packings[i];
 			var cv = makeVerifier(packingDecl, null, file);
+			if (prog.packings == null) prog.packings = Strings.newMap();
```
titzer, line 96: Can factor out the allocation from this loop.

aeneas/src/vst/Vst.v3
```diff
@@ -413,13 +413,14 @@ class SuperClause {
 type VstRepHint {
 	case Boxed;
 	case Unboxed;
-	case Packing(p: VstPackingExpr);
+	case Packing(p: List<VstPackingExpr>);
```
titzer, line 416: Probably best to use a `VstList`

aeneas/src/vst/Verifier.v3
```diff
@@ -93,6 +93,12 @@ class Verifier(compiler: Compiler, prog: Program) {
...
+			var name = packingDecl.name();
+			var prev = prog.packings[name];
+			if (prev != null) cv.errAtDecl(packingDecl).TypeRedefined(packingDecl.token);
```
titzer, line 99: Should be `PackingRedefined`

titzer (commented):
Still not convinced yet by the curly braces, can discuss over slack.

test/variants/parser/packing_concat01.v3
```diff
@@ -0,0 +1,2 @@
+//@parse
+packing MyScalar(a: 16, b: 16): 32 = a <<| b;
```
titzer, line 2: Can you add some more negative parsing tests? Those are usually the ones that reveal parser bugs.

aeneas/src/vst/Verifier.v3
```diff
@@ -886,29 +893,61 @@ class VstCompoundVerifier {
...
-			Tuple(list) => {
-				for (l = list.list; l != null; l = l.tail) {
-					checkPackingExpr(env, l.head, MAX_PACKING_WIDTH); // TODO: proper size
-				}
-			}
-			App(p, args) => { // TODO
-				// TODO: resolve and apply packing {p}
-				for (l = args.list; l != null; l = l.tail) {
-					checkPackingExpr(env, l.head, MAX_PACKING_WIDTH); // TODO: proper size
+				return byte.!(rep.length);
```
titzer, line 896: A naked cast here is somewhat dangerous (could crash the compiler, even if an error is reported elsewhere). Better to use a `view` or clamp.

aeneas/src/vst/Verifier.v3
```diff
@@ -886,29 +893,61 @@ class VstCompoundVerifier {
...
+			}
+			Concat(l) => {
+				var total: byte = 0;
+				for (e = l.list; e != null; e = e.tail) {
+					var s = inferPackingExpr(env, e.head);
+					if (s == MAX_PACKING_WIDTH || total == MAX_PACKING_WIDTH) total = MAX_PACKING_WIDTH;
```
titzer, line 902: IIUC this should be checking for the `total` doesn't exceed `MAX_PACKING_WIDTH`,  but it is only checking for equality, and will miss the case where the last expression goes over the limit.

aeneas/src/vst/Verifier.v3
```diff
@@ -886,29 +893,61 @@ class VstCompoundVerifier {
...
+					errAtRange(args.range()).ArityMismatch(p.ident.name.image, packingFn.pparams.length(), args.length());
+				} else {
+					var pp = packingFn.pparams.list;
+					for (l = args.list; l != null; l = l.tail) {
+						var param = pp.head, paramSize = byte.!(tryUnboxPositiveInt(param.size, -1));
+						checkPackingExpr(env, l.head, paramSize); // TODO: proper size
+						pp = pp.tail;
+					}
+				}
+				return byte.!(tryUnboxPositiveInt(packingFn.size, -1));
```
titzer, line 930: Similar here; we need to avoid casts so as not to crash the compiler for programs with type errors.

test/variants/seman/packing_app01.v3
```diff
@@ -0,0 +1,3 @@
+//@seman = TypeError @ 3:24
```
titzer, line 1: I would split this test into three; one that passes, one that fails on the first argument, and one that fails on the second argument, which makes sure that the compiler is performing type checking and it is checking all fields.

titzer (approved):
lgtm; can you bump the version number and I'll land this?

PR #226, opened 2024-03-23
[unbox] Fix variant equality to be case-by-case

btwj:
Variant equality is now done on a case-by-case basis. Previously, it did not correctly dispatch the variant equality method.

In the added test case, the old version would merge `B` and `C` together to become `#ref`, and then use `RefEq` instead, which is incorrect.

### DIFF
--- a/aeneas/src/ir/SsaNormalizer.v3
+++ b/aeneas/src/ir/SsaNormalizer.v3
@@ -691,8 +691,10 @@ class SsaRaNormalizer extends SsaRebuilder {
 	def normEqualOp(oldApp: SsaApplyOp, op: Operator) {
 		var tn = normTypeArg(op, 0);
 		var refs = genRefs(oldApp.inputs);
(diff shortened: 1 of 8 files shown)


aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -692,6 +692,7 @@ class SsaRaNormalizer extends SsaRebuilder {
 		var tn = normTypeArg(op, 0);
 		var refs = genRefs(oldApp.inputs);
 		if (!VariantNorm.?(tn) && V3.isVariant(tn.newType)) return map1(oldApp, normVariantEqual(oldApp, tn.newType, refs[0], refs[1]));
```
titzer, line 694: Might be worth doing a match on the `tn` type here since there are two cases now.

test/variants/ub_eq08.v3
```diff
@@ -0,0 +1,22 @@
+//@execute 0=true;1=false;2=true;3=false;4=true;5=false
```
titzer, line 1: Can you write a few more test cases, some smaller and some larger, and a couple that are polymorphic?

titzer (approved):
Generally looks good.

titzer (approved):
lgtm, thanks!

PR #231, opened 2024-03-27
[unbox] Unbox Variants with Scalar Sets

btwj:
This PR reworks how variants are solved. Instead of the previous approach using `unifyRepresentations`, the target provides a mapping from `Type` to a `Scalar.set`.
- Normalization of records now uses the SSA Interpreter, so logic isn't duplicated.

TODOs/Issues
- [ ] TODO: Heuristics for picking the scalar (if sets are equal, try using that first).
- [x] TODO: Don't regenerate IR for ClassAlloc for every record, do it once and cache.
- Issue: there is a problem with the JVM backend when doing `IntEq` with a `long` and `0:int` (generated by zeroConst). Had to create new boxed zeros and ones.


btwj:
@titzer I'm currently having trouble with the JVM target when it generates the heap image wrt closures. This happens when we have code pointers from a closure stored in a reference slot (`#ref`), see `ub_cell00.v3`. I'm also not that sure what the `JVMV3ClosureAdapterGen` is doing. Could you give me a couple pointers?

aeneas/src/ir/VariantSolver.v3
```diff
@@ -190,68 +217,138 @@ class VariantSolver(nc: NormalizerConfig, rn: ReachabilityNormalizer, verbose: b
...
+	}
 	private def unboxUsingTaggedVariantNorm(rc: RaClass) -> bool {
 		var vecT = Vector<Type>.new();
 		var parentUnboxed = rc.orig.boxing == Boxing.UNBOXED;
 
-		var scalars = ScalarSet.new(nc.UnifyRepresentations(rn.ra.compiler, rn.ra.prog, _, _));
-		var cases: List<(Array<VariantField>, Array<(int, int)>)>;
+		var numChildren = Lists.length(rc.children);
 
+		var cases = Array<(Array<VariantField>, Array<(int, int)>, Array<(int, int)>, Array<Type>)>.new(numChildren);
```
titzer, line 272: The `cases` array holds information for each child and is indexed in parallel to the children list. Does it make sense to instead put some state into `RaClass` that is only used here? It would simplify the iteration over children and the parallel array.

aeneas/src/jvm/JvmHeap.v3
```diff
@@ -100,7 +100,10 @@ class JvmHeap(jprog: JvmProgram) {
 			TUPLE => jprog.context.fail("normalization should remove all tuple values");
 			VARIANT,
 			CLASS,
-			ARRAY => visitRecord(Record.!(val), etype, ref, index);
+			ARRAY => match (val) {
```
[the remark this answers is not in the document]
btwj: As far as I can tell, I think this wouldn't have been hit before this change, because there was never an instance of JVM code pointer becoming cast to `AnyRef.TYPE`. Then, it would've fallen through to the `instrs.put(JvmHI_Value.new(val, etype))` case directly.

Now that we have this potential cast, it incorrectly tries to visit it as a record because the `kind` of `AnyRef.TYPE` is `CLASS`. I had to add this here to fix that so it wouldn't try to cast it as a record, instead relying on `val`'s type. Do you have a better workaround?

aeneas/src/ir/Normalization.v3
```diff
@@ -354,38 +341,30 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 		}
 		rc.normFields = fields.extract();
 	}
+	private def normRecordAsArray(rc: RaClass, oldRecord: Record, array: Array<Val>) {
+		var rfs = rc.fields;
+		for (i < rfs.length) {
+			var rf = rfs[i];
+			if (rf != null && rf.normIndex >= 0) {
```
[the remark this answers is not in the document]
btwj: Not sure of a good way to refactor this part without introducing closures - the only places where this logic appears is once in Normalization and twice in `VariantSolver`, but internal logic is slightly different. Wdyt?

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -1352,9 +1339,52 @@ class SsaRaNormalizer extends SsaRebuilder {
...
+		if (IntType.?(oldType) && IntType.?(newType)) return curBlock.opIntView(oldType, IntType.!(newType), i_scalar);
+		if (IntType.?(oldType)) {
+			match (newType) {
+				x: BoolType => return curBlock.opNotEqual(oldType, i_scalar, newGraph.nullConst(oldType));
+				x: FloatType => return curBlock.pure(if(x.is64, V3Op.newFloat64ViewI(oldType), V3Op.newFloat32ViewI(oldType)), [i_scalar]);
+			}
+		}
+		if (IntType.?(newType)) {
+			// XXX: Hack - Something goes wrong with the JVM backend if we use `oneConst()` and `zeroConst()` here
+			var one = if(IntType.!(newType).width > 32, newGraph.valConst(newType, Long.box(1)), newGraph.oneConst());
```
titzer, line 1353: You can use the `IntType.box()` method which chooses for you, and then use `SsaGraph.valConst`--it's a little less efficient, but cleaner here. This is probably where we should introduce the `IntViewBool`.

aeneas/src/os/Linux.v3
```diff
@@ -24,7 +24,24 @@ class LinuxTarget extends Target {
...
 			compiler.NormConfig.ArrayIndexType = it;
 		}
-		compiler.NormConfig.UnifyRepresentations = UnboxingUtil.makeMachUnifyRepresentations(getRuntime);
+		compiler.NormConfig.GetScalar = getScalar;
+	}
+	private def getScalar(compiler: Compiler, prog: Program, t: Type) -> Scalar.set {
+		if (space.addressWidth == 32) {
+			match (t) {
+				x: BoolType => return Scalar.B32 | Scalar.B64;
+				x: IntType => return Scalar.B32 | Scalar.B64; // XXX: Scalar.R64, once packed refs
```
titzer, line 33: You should check the size of the integer type here.

aeneas/src/wasm/WasmTarget.v3
```diff
@@ -89,7 +89,15 @@ class WasmTarget extends Target {
 	def configureCompiler(compiler: Compiler) {
 		compiler.Reachability = true;
 		compiler.NormConfig.setSignatureLimits(10000, if(CLOptions.WASM_MULTI_VALUE.val, 1000, 1));
-		compiler.NormConfig.UnifyRepresentations = UnboxingUtil.makeMachUnifyRepresentations(getRuntime);
+		compiler.NormConfig.GetScalar = getScalar;
+	}
+	private def getScalar(compiler: Compiler, prog: Program, t: Type) -> Scalar.set {
+		match (t) {
+			x: BoolType => return Scalar.B32 | Scalar.B64;
+			x: IntType => return if (x.width <= 32, Scalar.B32 | Scalar.B64, Scalar.B64); // XXX: Scalar.R64, once packed refs
```
titzer, line 97: You could write `Scalar.B64 | if(x.width <= 32, Scalar.B32)`

titzer (approved):
I think this is landable with some small changes. We can work on improving the data structures (e.g. numbering all the normalized fields of all cases so as to avoid a `HashMap`), but we can do that later.

aeneas/src/wasm/WasmTarget.v3
btwj: This doesn't seem to work. I think what's happening is `if(x.width <= 32, Scalar.B32)` is typechecked as having type `Scalar` instead of `Scalar.set`, so its default value in the `false` case will be `Scalar.B32` too (instead of the empty set).

aeneas/src/ir/SsaNormalizer.v3
btwj: Yeah, I'll do `IntViewBool` in a separate PR.

titzer (approved):
lgtm, can you bump the version number and I'll land?

PR #232, opened 2024-04-06
[packing] [WIP] Simple packing

btwj:
Progress on packing fields in a variant. 
- Introduces `IntRepType` that extends `IntType`. Allows for values of this type to undergo bitwise operations with the ordinary operators.

For now, I've introduced a `#packed` annotation to tell the compiler to pack a variant, while I'm still trying to figure out the best semantics for the `#packing` annotation.

TODOs
- [x] Share code between tagless and tagged variants so that packing works for both
- [ ] Add escape hatch for when solver is unable to find a solution - error out and report to user


aeneas/src/ir/Normalization.v3
```diff
@@ -24,11 +25,20 @@ class NormalizerConfig {
...
 		_ => return Scalar.Ref;
 	}
 }
+def defaultGetBitWidth(compiler: Compiler, prog: Program, t: Type) -> byte {
+	var target = compiler.target;
+	match (t) {
+		x: IntType => return x.width;
+		x: FloatType => return x.total_width;
+		x: BoolType => return 1;
+		_ => return 64;
```
titzer, line 39: I assume this is intended only for scalars?

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -1370,18 +1369,53 @@ class SsaRaNormalizer extends SsaRebuilder {
...
 				var idx = f.indexes[j];
-				result[idx] = genVariantScalarView(f.tn.at(j), vn.at(idx), ai_inputs[os + j]);
+				if (!f.isPacked()) result[idx] = genVariantScalarView(f.tn.at(j), vn.at(idx), ai_inputs[os + j]);
+				else result[idx] = genSetInterval(result[idx], ai_inputs[os + j], f.intervals[j], f.tn.at(j), IntRepType.!(vn.at(idx)));
 			}
 		}
 		return result;
 	}
+	def genExtractInterval(scalar: SsaInstr, interval: Interval, ft: IntRepType, tt: Type) -> SsaInstr {
+		if (interval.start > 0) scalar = curBlock.addApplyF(ft.opShr(), [scalar, newGraph.intConst(interval.start)], Fact.O_PURE | Fact.O_NO_SHIFT_CHECK);
```
titzer, line 1394: `Fact.O_PURE | Fact.O_NO_SHIFT_CHECK` occurs several times; might be worth factoring out to `Facts`.

aeneas/src/ir/VariantSolver.v3
```diff
@@ -83,6 +101,197 @@ def ON_STACK = -1;
...
+			return solvePacking();
+		}
+		if (curField >= normFields[curCase].length) {
+			// move on to next case
+			curUsed = Vector<byte>.new();
+			for (i < curRep.length) curUsed.put(0);
+			return tryRepresentationForField(curCase + 1, 0);
+		}
+		
+		var sc = vnorm.nc.GetScalar(vnorm.rn.ra.compiler, vnorm.rn.ra.prog, normFields[curCase][curField]);
```
titzer, line 137: The `VariantNormalizer` should probably have a method that hides its configuration and the compiler and program arguments.

aeneas/src/ir/VariantSolver.v3
```diff
@@ -83,6 +101,197 @@ def ON_STACK = -1;
 def EMPTY_FIELD = VariantField.new(null, []);
 def NO_FIELDS = Array<VariantField>.new(0);
 
+class VariantSolver(vnorm: VariantNormalizer, usePacking: bool, useTagScalar: bool) {
+	var normFields: Array<Array<Type>>;
+
+	var fields: Vector<CaseField>; // an ordering of fields to solve for during packing
+
+	// final solutions
+	var scalarAsns: HashMap<CaseField, int>; // field location -> scalar idx
```
titzer, line 110: Any reason not to combine the scalar assignment and interval assignment into one hashmap?

aeneas/src/ir/VariantSolver.v3
```diff
@@ -83,6 +101,197 @@ def ON_STACK = -1;
...
+		buf.ln();
+		for (i < state.length) {
+			buf.put1("%q\n", state[i].render);
+		}
+		return buf;
+	}
+	def reset() {
+		scalarAsns = HashMap<CaseField, int>.new(CaseField.hash, CaseField.==);
+		intervalAsns = HashMap<CaseField, Interval>.new(CaseField.hash, CaseField.==);
+		curRep = Vector<Scalar.set>.new();
```
titzer, line 272: You can use `.resize(0)` here.

aeneas/src/ir/VariantSolver.v3
```diff
@@ -83,6 +101,197 @@ def ON_STACK = -1;
...
+		curUsed = Vector<byte>.new();
+		fields = Vector<CaseField>.new();
+	}
+	private def getTagLength(numCases: int) -> byte {
```
titzer, line 276: You should be able to use `Ints.log` here.

aeneas/src/ir/VariantSolver.v3
```diff
@@ -151,79 +360,65 @@ class VariantNormalizer(nc: NormalizerConfig, rn: ReachabilityNormalizer, verbos
 	}
 	private def unboxUsingTaglessVariantNorm(rc: RaClass) {
 		var ofs = rc.orig.fields;
```
titzer, line 362: This is unused.

aeneas/src/ir/VariantSolver.v3
```diff
@@ -151,79 +360,65 @@ class VariantNormalizer(nc: NormalizerConfig, rn: ReachabilityNormalizer, verbos
...
+
+		vn.fieldRanges = rc.fieldRangesO;
+		vn.vecO = rc.vecO;
+		rc.variantNorm = vn;
+		setVariantFields(rc, 0, solver, shouldPack);
+
+		if (verbose) Terminal.put1("variant norm %q\n", rc.variantNorm.render);
+	}
+	private def getNormFieldsForCase(rc: RaClass) -> Array<Type> {
+		var ofs = rc.orig.fields;
```
titzer, line 380: I think this while be cleaner if you iterate over `rc.fields`

aeneas/src/ir/VariantSolver.v3
```diff
@@ -151,79 +360,65 @@ class VariantNormalizer(nc: NormalizerConfig, rn: ReachabilityNormalizer, verbos
...
-				indexes = Array<int>.new(tn.size);
-				for (j < indexes.length) indexes[j] = normStart + j;
-				fields[i] = VariantField.new(tn, indexes);
+				tn.addTo(fieldTypes);
+				fields[i] = VariantField.new(tn, Array<int>.new(tn.size));
 			} else fields[i] = EMPTY_FIELD;
-			
+
+			var origStart = vecO.length;
 			var fieldType = ofs[i].fieldType.substitute(V3.getTypeArgs(rc.oldType));
```
titzer, line 399: If you have the `RaField` you shouldn't need to do a type substitution. If you're doing a type substitution then it's usually a sign you should be working with `Ra..` entities instead. E.g. the type you end up with might not have been encountered (and processed) yet.

aeneas/src/ir/VariantSolver.v3
```diff
@@ -83,6 +101,197 @@ def ON_STACK = -1;
...
+		// TODO - this is where unassigned bits will get assigned
+	}
+	private def checkDistinguishable() -> bool {
+		if (useTagScalar) return true;
+		if (normFields.length <= 1) return true; // single case is always distinguishable
+		var result = tryExplicitTaggingHeuristic();
+		// TODO: try difficult tagging
+		return result;
+	}
+	private def solveField(idx: int) -> bool {
```
titzer, line 208: I'm starting to wonder if this would be cleaner to just do an iterative solution, e.g. by sorting and using first-first in decreasing order. Also, we're repeatedly getting the width of a field; seems like that should be cached internally in the solver.

aeneas/src/ir/VariantSolver.v3
```diff
@@ -83,6 +101,197 @@ def ON_STACK = -1;
...
+			for (i < curRep.length) curUsed.put(0);
+			return tryRepresentationForField(curCase + 1, 0);
+		}
+		
+		var sc = vnorm.nc.GetScalar(vnorm.rn.ra.compiler, vnorm.rn.ra.prog, normFields[curCase][curField]);
+		var bw = vnorm.nc.GetBitWidth(vnorm.rn.ra.compiler, vnorm.rn.ra.prog, normFields[curCase][curField]);
+		var none: Scalar.set;
+
+		for (i < curRep.length) {
+			if (usePacking && curUsed[i] + bw > 64) continue;
```
titzer, line 142: 64 is an arbitrary limit...it needs to be factored out somewhere, e.g. the variant configuration.

aeneas/src/ir/VariantSolver.v3
```diff
@@ -83,6 +101,197 @@ def ON_STACK = -1;
...
+
+	var curRep: Vector<Scalar.set>;
+	var curUsed: Vector<byte>;
+	
+	// Perform recursive backtracking on the potential representation.
+	// With each scalar in each case, we have several choices to make:
+	// 1. Pack this scalar with an existing scalar
+	// 2. Append this scalar as a new scalar.
+	// We can only determine distinguishability after all the scalars have been assigned. Heuristics will help speed this part up.
+	private def tryRepresentationForField(curCase: int, curField: int) -> bool {
```
titzer, line 125: It's hard to follow how this works because this an internal part of the recursion; the overall recursive algorithm should come first and be documented; then each individual part should be documented, along with the arguments.

aeneas/src/ir/VariantSolver.v3
btwj: I think `Ints.log` returns the floor of the log (see `IntsTest.v3`), but I need the ceil here. Is there another method I can use?

aeneas/src/ir/VariantSolver.v3
btwj: No reason, I'll combine them.

aeneas/src/ir/VariantSolver.v3
btwj: I'll take a look at this and see if I can safely remove this. I think this was carried over from the original implementation of normalization using `DataNorm`s, but I don't know if it's doing anything at this point.

btwj: We need this type substitution here in computing `vecO`, since `RaField`'s aren't generated for fields never encountered, but `vecO` is to help us normalize allocations of the class, and the arguments for unused fields aren't eliminated.

aeneas/src/ir/VariantSolver.v3
btwj: What do you mean here by first-first in decreasing order? I can try the iterative method, but it would be tricky to do the recursive backtracking iteratively (i.e. trying a new interval to slot a field in if the original interval doesn't work).

titzer: Sorry, I meant "first-fit" in decreasing order. But I guess this is landable in its current form.

aeneas/src/ir/VariantSolver.v3
titzer: It's probably worthwhile having a `log` that does the ceiling. Can do in a followup.

titzer (approved):
I think we can land this and work on further improvements in followups.

titzer:
Can you bump the version number and I'll land this?

PR #238, opened 2024-04-18
[packing] Packing Attempt #2

btwj:
Mostly the same as the old implementation, with:
- No use of `SsaBlockSplit` for the case where a `bool` is viewed as an `int`. `SsaBlockSplit` breaks for constant conditions when used to rewrite SSA. I can't quite figure out how to modify `SsaBlockSplit` to handle the full generality of an arbitrary number of cases, so I just didn't use it in this case. All other instances of `SsaBlockSplit` are not constants, which is why this problem never arose. Regardless, we'll probably want a custom `BoolViewI` operator here.
- I added an optimization to remove some classes of type subsumption. Basically, it looks if a scalar is only used to represent a certain type and is never packed. If it is, then it uses that type in the representation instead of weaking to a `u32[B32]` or `u64[B64]` etc.
- Refactoring to package the solution up as a `VariantSolution` (which will have an associated `score` function to be implemented). For now, it still just outputs the first solution it finds.
- Prettier printing of variant norms and unboxing with the `-print-packing` flag:
image


titzer (approved):
LGTM

PR #241, opened 2024-04-23
Fix bug with unboxed refs

btwj:
We should not be using `u32[R32]/u64[R64]` at all while packed refs are not supported.
Fixes bug so `type Operand #unboxed #packed` can bootstrap.

### DIFF
--- a/aeneas/src/ir/VariantSolver.v3
+++ b/aeneas/src/ir/VariantSolver.v3
@@ -17,7 +17,10 @@ def getScalarFromSet(s: Scalar.set) -> Scalar {
 	return Scalar.B32; // should be unreachable
 }
 def getTypeFromScalarSet(s: Scalar.set, usePacking: bool) -> Type {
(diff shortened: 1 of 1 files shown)


aeneas/src/ir/VariantSolver.v3
```diff
@@ -17,7 +17,10 @@ def getScalarFromSet(s: Scalar.set) -> Scalar {
 	return Scalar.B32; // should be unreachable
 }
 def getTypeFromScalarSet(s: Scalar.set, usePacking: bool) -> Type {
-	for (i in s) return if(usePacking, IntRep.getType(i), i.ty);
+	for (i in s) {
+		if (i.tag >= Scalar.R32.tag) return AnyRef.TYPE;
```
titzer, line 21: Checking the tag here is somewhat brittle; I think a `match` would be best here.

titzer:
Fixed this in a separate CL. If you have a testcase that fails, please share.

PR #252, opened 2024-06-16
[net] Minor refactorings to make Net.v3 more Virgilic

titzer:
Creating a PR to get input from @k-sareen on some minor refactorings to make the `net` code more Virgilic.

### DIFF
--- a/lib/net/x86-64-linux/Net.v3
+++ b/lib/net/x86-64-linux/Net.v3
@@ -23,7 +23,7 @@ enum NetError(desc: string) {
 	WouldBlock("Operation would block"),
 	InvalidInput("Invalid input parameter"),
 	InvalidProtocol("Using incorrect protocol"),
-	InvalidSockAddr("Invalid sockaddr size"),
+	InvalidSockAddress("Invalid sockaddr size"),
 	InvalidRange("Invalid range provided"),
 	TimedOut("Timed out"),
 	ResourceBusy("Resource busy"),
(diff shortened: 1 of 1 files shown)


lib/net/x86-64-linux/Net.v3
```diff
@@ -455,6 +423,15 @@ class Socket {
 		isConnected = false;
 		isListening = false;
 	}
+	// Overridden in subclasses to get the local address.
+	def getLocalAddre() -> InetAddress;
```
k-sareen, line 427:
```suggestion
	def getLocalAddress() -> InetAddress;
```

k-sareen (commented):
I need to pull and try writing some code against it (will do later tonight), but so far it looks okay other than the typo.

This code was written before layouts were merged so we could probably simplify the `toSockAddr` function as well which is yucky code to construct the C socket structs for interfacing with the kernel, but I'm not so familiar with how layouts work as I haven't sat down and implemented something with it (yet).

titzer:
> This code was written before layouts were merged

Indeed, I thought of going ahead and doing this, but alas, it uses big-endian integers, which are on the TODO list for layouts. I was not sure how client code uses serialized addresses, but another possible design is to have methods that write the serialized address out, and client code supplies an allocated for `Range<byte>`, which could be a method like the `acquire` method of a `DataWriter`.

k-sareen (approved):
LGTM. Thanks!

k-sareen:
> Indeed, I thought of going ahead and doing this, but alas, it uses big-endian integers, which are on the TODO list for layouts.

Ah yeah the network-byte order thing is annoying. Tbh I am not sure why it was designed in such a way, but here we are.

> I was not sure how client code uses serialized addresses, but another possible design is to have methods that write the serialized address out, and client code supplies an allocated for `Range<byte>`, which could be a method like the `acquire` method of a `DataWriter`.

I'm not sure I follow? Are you asking how code that uses this library is supposed to create new addresses? You would just need to define an `InetAddress` (so for example, `def inetAddr = InetAddress.Ipv4(10,0,10,6);`) and then can use `inetAddr` anywhere you want to create a socket.

PR #253, opened 2024-06-17
Add support for lock prefix

khagankhan:
This adds support for the lock prefix to be used when writing is `asm`.


lib/asm/x86-64/X86_64Assembler.v3
```diff
@@ -452,6 +452,7 @@ class X86_64Assembler(w: DataWriter, OP_REX: byte) {
 	}
 	def cdq() -> this { emitb(0x99); }
 	def cqo() -> this { emitbb(REX_BYTE | REX_W, 0x99); }
+	def lock() -> this { emitb(0xF0); }
```
titzer, line 455: Can you add this to the `X86Assembler` as well? (I assume it's the same encoding).

khagankhan: Done! (Yes, I've tested, and  it appears to be the same encoding).

titzer (approved):

PR #254, opened 2024-06-18
fix actions/checkout@v4

whonion:


titzer:
Can you explain briefly what the new version of the checkout action offers?

kalwalt:
@titzer it fix these warnings: `Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v2. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.`
you can check in the actions tab, for example:
https://github.com/titzer/virgil/actions/runs/9554786647
you can check them in the Annotations.

PR #258, opened 2024-06-19
Add Extism PDK

gmlewis:
This PR represents my attempt at writing an [Extism PDK](https://extism.org/docs/concepts/pdk/) for the Virgil programming language.

As a Virgil newbie, I may not have done things in an idiomatic way, and am open to learning and recommendations on making this better.


apps/Extism/count-vowels/CountVowels.v3
```diff
@@ -0,0 +1,77 @@
...
+// string and keeps a running total (over multiple iterations)
+// in the host's "total" var.
+// It sends the JSON `VowelReport` to the host via its output data channel.
+// It returns 0 to the host on success.
+def count_vowels() -> int {
+  def input = Host.inputString();
+  //
+  def vowels = getVowels();
+  vowelsArr = Array<byte>.!(vowels);
+  def count = Arrays.filter(input, filterFunc).length;
```
titzer, line 56: You can do `Arrays.contains(vowels, _)` (i.e. a partial application) here.

apps/Extism/count-vowels/CountVowels.v3
```diff
@@ -0,0 +1,77 @@
+// `defaultVowels` represents the default set of vowels
```
titzer, line 1: Virgil style is generally to use tabs instead of spaces for indentation.

apps/Extism/count-vowels/CountVowels.v3
```diff
@@ -0,0 +1,77 @@
...
+// Unused.
+def main() {
+}
+
+export count_vowels;
+
+// Hack to compile to wasm: provide Virgil identifiers that cannot be found:
+component System {
+  def error(s1: string, s2: string) {}
+  def fileClose(file: int) {}
```
titzer, line 75: I've now gotten rid of the usage of `fileClose` and `fileWriteK` in `lib/util`, so you can remove these two.

lib/pdk/Header.v3
```diff
@@ -0,0 +1,49 @@
+// `Header` represents an HTTP Request header.
+// Multiple values for a single key are not deduped.
+class Header {
+  var m: HashMap<string, string>;
```
titzer, line 4: If `m` is immutable, you could do the more "Virgilic" way of having `def m = Strings.newMap<string>();` right in the field declaration. It's generally good to have at least full-word field names, so `map` might be a better name.

lib/pdk/Host.v3
```diff
@@ -0,0 +1,99 @@
+// `Host` represents functions used to interact with the Extism host SDK.
+component Host {
+  // `input` returns a sequence of bytes from the host.
+  def input() -> Array<byte> {
+    def length = Extism.input_length();
+    def value = Array<byte>.new(int.view(length));
+    for (j = 0; j < length; j = j + 1) {
```
titzer, line 7: You can use the `for (j < length)` short form here.

lib/pdk/Host.v3
```diff
@@ -0,0 +1,99 @@
...
+
+  // `inputString` returns a string from the host.
+  def inputString() -> string {
+    return string.!(input());
+  }
+
+  // `outputBytesToMemory` writes the bytes to a Memory buffer on the host.
+  private def outputBytesToMemory(b: Array<byte>) -> Memory {
+    def offset = Extism.alloc(b.length);
+    for (i = 0; i < b.length; i = i + 1) {
```
titzer, line 21: Here, too.

lib/pdk/Http.v3
```diff
@@ -0,0 +1,57 @@
+// `Request` represents an HTTP request made by the Extism host.
+class Request {
```
titzer, line 2: If you like you can shorten this up with:

```
class Request(var method: Method, var header: Header, var url: string) { ...
```

If they're intended to be immutable, then you can drop the `var`.

lib/pdk/Memory.v3
```diff
@@ -0,0 +1,38 @@
+// `Memory` represents memory allocated by (and shared with) the host.
+class Memory {
```
titzer, line 2: Could be nicely shortened to:

```
class Memory(offset: i64, length: i64) { ...
```

lib/pdk/Method.v3
```diff
@@ -0,0 +1,39 @@
...
+  OPTIONS
+  // The TRACE method performs a message loop-back test along the path to the target resource.
+  TRACE
+  // The PATCH method applies partial modifications to a resource.
+  PATCH
+}
+
+component Methods {
+  // TODO - what is the idiomatic way to convert an enum to a string?
+  def name(m: Method) -> string {
```
titzer, line 26: Enums come with a `.name` field which matches the declaration, so you can get away without this method at all :-)

lib/util/Map.v3
```diff
@@ -110,6 +110,36 @@ class HashMap<K, V> extends PartialMap<K, V> {
 			}
 		}
 	}
+	// Number of key/value pairs in HashMap (expensive operation - cache the value).
```
titzer, line 113: I think I'd rather keep this out of the main `HashMap` implementation for now; can you do this with an `apply()` in your code for now?

lib/util/Map.v3
```diff
@@ -110,6 +110,36 @@ class HashMap<K, V> extends PartialMap<K, V> {
...
+		}
+		// two or more entries
+		var r = 0;
+		for (b in table) {
+			for (bucket = b; bucket != null; bucket = bucket.next) { r++; }
+		}
+    return r;
+	}
+	// Apply {func} to every (key, value) pair in this map with index.
+	def applyWithIndex(func: (K, V, int) -> void) {
```
titzer, line 128: Similarly here, I think this is enough of a special case I'd rather leave it to user code for now.

apps/Extism/count-vowels/CountVowels.v3
gmlewis: A standard formatter like Go or MoonBit would be fantastic. 
I'm sure it's probably on the roadmap.

gmlewis:
Thank you for the feedback, @titzer !
I believe I have addressed it all.
Please take another look when you get a chance.

titzer (approved):
lgtm, thanks!

Issue #77, opened 2022-08-20
UTF-8 string literals


srackham:
Does Virgil support UTF-8 string literals?

The documentation suggests it does: 
https://github.com/titzer/virgil/blob/3038dead280099b736f312e2b091b053cb0cfbf7/doc/lib-issues.txt#L116

Here I've inserted the copyright character in a string literal:

```
$ cat hello.v3    
def main() {
        System.puts("Hello World ©\n");
}

$ virgil run tmp/hello.v3
[tmp/hello.v3 @ 2:21] ParseError: invalid string literal
        System.puts("Hello World ©\n");
                    ^
```
Hex byte values work though:

```
$ cat hello.v3
def main() {
        System.puts("Hello World \xC2\xA9\n");
}

$ virgil run hello.v3
Hello World ©
```

titzer:
You're right, that's a bug. It should handle UTF-8 in string literals, but it does not yet.

I was planning on improving the support for unicode by changing the `string` type (currently an alias for `Array<byte>`), but this is something that could maybe supported by just allowing the UTF-8 representation through.

srackham:
Thanks.

A workaround is to convert UTF-8 strings to hex byte values with, for example:

```sh
$ echo -n "Hello World ©" | od -A n -t x1 | tr -d '\n' | sed 's/ /\\x/g'
\x48\x65\x6c\x6c\x6f\x20\x57\x6f\x72\x6c\x64\x20\xc2\xa9
```

PR #259, opened 2024-06-27
Fix issue #77 Support for UTF-8 string literals

yoyocat:


lib/util/Strings.v3
```diff
@@ -133,16 +133,22 @@ component Strings {
...
-			if (ch < ' ' || ch > 127) return (pos - i, null);
+			if (ch < ' ') return (pos - i, null);
 			if (ch == '\"') return (1 + i - pos, buf.toString());
 			if (ch == '\\') {
 				var p = Chars.parseEscape(a, i + 1);
 				if (p.0 <= 0) return (pos - i + p.0, null);
 				else buf.putc(p.1);
 				i = i + p.0;
 			} else {
+				if (ch > 127 && !checked) {
```
titzer, line 147: As I understand this code, upon the first non-ascii character, this checks the rest of the input buffer (i.e. until the end, which could be past the closing `"`) and then proceeds to just add the non-ascii byte to the buffer. It only does this once, but I still think that's not quite right, because after adding the first byte of the UTF-8 encoded character, it will just go back through the loop with the next byte. I think this should probably decode the whole codepoint and add those bytes to the buffer, then continue in the loop.

apps/HelloWorld/HelloWorld.v3
```diff
@@ -1,3 +1,3 @@
 def main() {
-	System.puts("Hello World!\n");
+	System.puts("Hello World!\n你好，世界!\n");
```
titzer, line 2: I think this could use a new battery of tests in `test/lib` and `test/core`.

titzer (commented):
Thanks for taking a stab at this! There are a few more things to make this robust.

lib/util/Strings.v3
yoyocat: OK. I'll fix it. I wrongly assume that `max` is the end of the string.

yoyocat:
👍 💯 🥇

PR #260, opened 2024-06-28
Fix issue #77 Support for UTF-8 string literals

yoyocat:


lib/util/Utf8.v3
```diff
@@ -46,47 +48,55 @@ component Utf8 {
 		if (codepoint >= 0xD800 && codepoint <= 0xDFFF) return false; // UTF-16 surrogates
 		return true;
 	}
+	// Decode a UTF-8 sequences of bytes to one Unicode codepoint. 
+	// Returns (byte offset of the char, codepoint).
+	def decode1(str: Array<byte>, start: int, end: int) -> (int, u32) {
```
titzer, line 53: Accepting a `Range<byte>` as input will make this a little more general and robust.

Also, what about returning `(-1, 0)` in case of error, instead of an invalid codepoint? That would be similar to how `Ints.parseDecimal()` works.

test/core/parser/str02.v3
```diff
@@ -7,4 +7,7 @@ class A {
 	var j: string = "`~!@#$%^&'*()-_=+";
 	var k: string = "[{]}|;:,<.>/?";
 	var l: string = "\r\t\n\"\'\\\x4C";
+	var m: string = "--你好，世界！--";
```
titzer, line 10: Can you split this test out to a separate file (so tests are additive)?

titzer (commented):
Thanks for the update. Only a couple of minor things, then we can land this.

lib/util/Utf8.v3
yoyocat: > Also, what about returning (-1, 0) in case of error, instead of an invalid codepoint? That would be similar to how Ints.parseDecimal() works.

We need to return the byte offset of the invalid character, so there are several error cases. Only `-1` is not enough. What about changing the spec of the function as below?
```
def decode1(str: Array<byte>, r: Range<byte>, f: u32 -> void) -> (int, bool)
```
We return a bool value to indicate success or failure.

test/core/parser/str02.v3
yoyocat: Sure.

titzer:
Just FYI, the failure in CI is because the `test/lib` test is run with both the stable compiler and the bootstrap compiler, and the stable compiler doesn't support UTF-8 yet. (That will require a stable rev). So it's enough to add tests in, e.g. `test/core` for this.

lib/util/Utf8.v3
yoyocat: @titzer Splited the tests. Should I change this too?

titzer: Sorry for the late reply. If this function instead accepts a `Range<byte>`, I think its OK for this function to just return at `(u32, bool)` pair, because the caller can either assume byte 0 is wrong, or if finer granularity is needed, figure out which byte is wrong.

titzer (approved):
lgtm, thanks!

PR #261, opened 2024-07-04
Add exchange instructions to x86-64Assembler

khagankhan:
Adds support for exchange instructions in `x86-64` that can be used  for atomic stores and other purposes.


khagankhan:
Do we need `xchgq_m_r` and `xchgd_m_r` or can we  (I think we can) simply use `asm.d.xchg_m_r` and `asm.q.xchg_m_r`?

titzer:
> Do we need `xchgq_m_r` and `xchgd_m_r` or can we (I think we can) simply use `asm.d.xchg_m_r` and `asm.q.xchg_m_r`?

I think I prefer the explicit suffixes here.

PR #267, opened 2024-07-24
Fix bugs with `-unbox-variants`

btwj:
Fixes two separate bugs with variant unboxing uncovered by the `-unbox-variants` flag:
- Recursive tests are more exhaustive: previously, it would attempt to unbox variants even if cases referred to their sibling types.
- We were saving space by caching the ClassAlloc constructor in `rc.orig.methods[0]`. This is not correct, because multiple `RaClass`es can have the same `rc.orig`.

### DIFF
--- a/aeneas/src/ir/VariantSolver.v3
+++ b/aeneas/src/ir/VariantSolver.v3
@@ -97,7 +97,7 @@ class VariantField(tn: TypeNorm, indexes: Array<int>) {
 	}
 }
 
-def ON_STACK = -1, IS_RECURSIVE = 1, NOT_RECURSIVE = 2;
+def ON_STACK = -1, IS_RECURSIVE = 2, NOT_RECURSIVE = 1;
 def EMPTY_FIELD = VariantField.new(null, []);
 def NO_FIELDS = Array<VariantField>.new(0);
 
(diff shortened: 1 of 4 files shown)


aeneas/src/ir/VariantSolver.v3
```diff
@@ -385,31 +385,37 @@ class VariantNormalizer(nc: NormalizerConfig, rn: ReachabilityNormalizer, verbos
 	// Entrypoint into representation selection. Recursively tries to unbox and pack this variant.
 	// Returns either a {VariantNorm} for an unboxed variant or a simple {TypeNorm} for a boxed variant.
 	def normVariant(t: Type, rc: RaClass) -> TypeNorm {
-		if (rc.variantNorm != null) return rc.variantNorm; // already done
+		if (rc.variantNorm != null)	return rc.variantNorm; // already done
```
titzer, line 388: Spurious tab?

titzer (commented):
Thanks for looking into this!

aeneas/src/ir/VariantSolver.v3
```diff
@@ -385,31 +385,36 @@ class VariantNormalizer(nc: NormalizerConfig, rn: ReachabilityNormalizer, verbos
 	// Entrypoint into representation selection. Recursively tries to unbox and pack this variant.
 	// Returns either a {VariantNorm} for an unboxed variant or a simple {TypeNorm} for a boxed variant.
 	def normVariant(t: Type, rc: RaClass) -> TypeNorm {
+		Terminal.put2("normVariant rc=%q rec=%d\n", rc.oldType.render, rc.recursive);
```
titzer, line 388: Some leftover tracing code here.

titzer (commented):
Mostly rubber-stamped. Can you add some tests that would have/were failing?

test/variants/ub_rec06.v3
```diff
@@ -0,0 +1,28 @@
+//@execute 0=0; 1=5; 2=10; 3=120
+
+type T<E> #unboxed {
+	case Leaf(x: E);
+	case One(a: T<E>.Leaf, b: T<E>.Leaf);
+	case Two(a: T<E>.One, b: T<E>.One);
+}
+
+def main(a: int) -> int {
```
titzer, line 9: Can you add a third test which has two different instantiations of this polymorphic type, with one instantiation being a tuple or other type that normalizes to more than one scalar?

titzer (approved):
lgtm; can you bump the version number and add another test or two?

test/variants/ub_rec06.v3
titzer: I'll merge this for now; I can do the bump and the test.

PR #269, opened 2024-08-09
Fix jvm namespace clash bug

thelongmarch-azx:
When the compiler emits jvm/jar target, only virgil class names are considered when creating the jvm class. This means private virgil classes in different files having the same name will lead to a jvm runtime error, especially when the signatures of their constructors are different. This PR incorporates the virgil type uid into the jvm class name, thus avoiding this name clash issue.


test/namespace/lib.v3
```diff
@@ -0,0 +1,9 @@
+class Util {
```
titzer, line 1: I think we can test this by just having another application in `apps/`

aeneas/src/jvm/JvmRep.v3
```diff
@@ -56,7 +56,8 @@ class JvmTypeReps(jprog: JvmProgram) {
 	}
 	private def makeJvmClass(superRep: JvmType, utype: Type) -> JvmType {
 		var ic = jprog.prog.ir.getIrClass(utype);
-		var className = jprog.mangleTypeName("V3C_", utype);
+		var base_name = if (ic == null, "V3C_", StringBuilder.new().put1("V3C_%d", ic.uid).toString());
```
titzer, line 59: Names are useful when crashing on the JVM. I think this would be better if `mangleTypeName` just appends the UID if the class was private.

titzer (commented):
Thanks for looking into this!

aeneas/src/jvm/JvmRep.v3
thelongmarch-azx: It seems like a type's `isPrivate` info is only accessible at the Verifier, and takes quite a lot of code changes to be made available in the backend, so I think it'll be safer to have `mangleTypeName` append the uid regardless of `isPrivate`

titzer: You should be able to get the privacy like this:

```
var isPrivate = if(ClassType.?(utype), ClassType.!(utype).classDecl.isPrivate);
```

test/namespace/lib.v3
thelongmarch-azx: I added it to `bench/`

titzer (approved):
lgtm, thanks!

PR #283, opened 2024-09-28
Arm64 Assembler Updates

YonahGoldberg:
- Added some more data processing instructions
- Function type suffixes are more explicit
- More careful use of signed/unsigned integers


lib/asm/arm64/Arm64Assembler.v3
```diff
@@ -4,6 +4,25 @@
 // Assembler for Arm64 Architecture
 // See https://developer.arm.com/documentation/ddi0487/latest/ for instruction and opcode information
 
+/* Note: Why use an unsigned int vs a signed int?
```
titzer, line 7: While this comment is nice context, I think it doesn't really belong here and reads somewhat oddly (e.g. is first person). It could be replaced with 

"Signedness of integers follows from context, and unsigned integers are used in sign-agnostic cases".

titzer (approved):
Other than the comment change, lgtm.

PR #286, opened 2024-10-02
Arm64 Backend Updates

YonahGoldberg:
Lots of updates here.

- Added an Arm64MacroAssembler to allow for operations that emit multiple assembly instructions. This is useful often in the backend.
- The Arm64MacroAssembler also has operations that take addresses as operands and handles patching internally.
- SsaArm64Gen is reworked so that it never calls iN.view on an integer type. I did this initially because Arm64 doesn't support large immediates in instructions, so it was easiest to just "view" the integer as a smaller type. This is dangerous and could cause unexpected bugs later on because views can truncate. SsaArm64Gen now makes the type of integers explicit in the MachIR by using Box<iN>. This allows for the MachIR to be closer to a mirror of the emitted assembly.
- Added a second scratch register for the Arm64 backend. This is explained in a comment in Arm64RegSet.v3.
- The only time the Arm64 backend calls view on an immediate is when patching addresses, which is a difficult corner case that I'll handle later.
- Added pretty printing for the Arm64 MachIR, which I'll continue updating as I add more support. This includes a small change to other MachIRs, where immediates always show they're type when printing. Below is an example.
Screenshot 2024-10-02 at 12 48 32 PM

### DIFF
--- a/aeneas/src/arm64/SsaArm64Gen.v3
+++ b/aeneas/src/arm64/SsaArm64Gen.v3
@@ -106,46 +106,6 @@ class SsaArm64Gen extends SsaMachGen {
 	// Overidden Architecture Specific Routines
 	// ========================================
 
-	// Given an integer x of type iW, use an immediate of type iN to represent it.
-	// Precondition: x can be represented as an iN
-	def useFixedInt<N, W>(x: W) {
-		op(Operand.Immediate(Box<N>.new(N.!(x))));
-	}
-	// Try to use an immediate T for SSA instruction I. Arm64 has weird immediate sizes for
-	// some instructions, so T can be any width integer. If i fits in an integer of size T,
-	// use an integer operand with Box<T> and return true. Otherwise, return false.
-	def tryUseFixedInt<N>(i: SsaInstr) -> bool {
-		if (i == null) { useInt(0); return true; }
-		if (SsaConst.?(i)) {
-			var val = SsaConst.!(i).val;
-			match (val) {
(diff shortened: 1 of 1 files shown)


aeneas/src/arm64/SsaArm64Gen.v3
```diff
@@ -129,87 +224,288 @@ class SsaArm64Gen extends SsaMachGen {
...
+	// ========================================
+	// Assembling Helpers
+	// ========================================	
+
+	def assemble_none(opcode: int, a: Array<Operand>) {
+		match (getCode(opcode)) {
+			I_BL => {
+				var target: bool, outgoing: MachCallConv, livepoint = -1;
+				for (o in a) {
+					match (o) {
```
titzer, line 274: It seems like `BL` doesn't really fit under the "none" addressing mode. If you're switching over operands, then it's better to encode that into the addressing mode.

aeneas/src/v3/V3.v3
```diff
@@ -152,6 +152,14 @@ component V3 {
 			return buf;
 		}
 		match (val) {
+			x: Box<u1> => buf.putd(x.val);
```
titzer, line 155: I've been trying to move the compiler away from using `Box<int>`, `Box<long>`, etc, to having dedicated value classes, instead of a parameterized class. I think it'd be less error prone to try to use `Box` to enforce the bitwidth of immediates. Therefore I think it'd be better to use `Box<int>` for now for all integer-sized immediates.

PR #288, opened 2024-10-06
Added search functions in `Ranges.v3`

linxuanm:
Added various binary search functions from the optimization PR in Wizard to `Ranges.v3`.

TODO: add tests in `RangesTest.v3`

### DIFF
--- a/lib/util/Ranges.v3
+++ b/lib/util/Ranges.v3
@@ -76,4 +76,40 @@ component Ranges {
 		}
 		return i;
 	}
+	// Performs binary search and returns the index.
+	def binarySearch<T>(range: Range<T>, val: T, lt: (T, T) -> bool) -> int {
+    var l = 0, r = range.length;
+    while (l < r) {
+      var mid = (l + r) >> 1;
+      var mid_val = range[mid];
+      if (lt(val, mid_val)) r = mid - 1;
+      else if (lt(mid_val, val)) l = mid + 1;
+      else return mid;
+    }
+    return -1;
+  }
+  // Returns the *first* element that is *not less* than `val` in a sorted range.
+  def lowerbound<T>(range: Range<T>, val: T, lt: (T, T) -> bool) -> int {
+    var start = 0;
+    var count = range.length;
+    while (count > 0) {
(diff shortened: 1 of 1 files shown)


lib/util/Ranges.v3
```diff
@@ -76,4 +76,40 @@ component Ranges {
 		}
 		return i;
 	}
+	// Performs binary search and returns the index.
```
titzer, line 79: This code should follow the tab-per-indent level style of the surrounding Virgil code.

lib/util/Ranges.v3
```diff
@@ -78,38 +78,38 @@ component Ranges {
...
+		return start;
+	} 
+	// Returns the *first* element that is *greater* than `val` in a sorted range.
+	def upperbound<T>(range: Range<T>, val: T, lt: (T, T) -> bool) -> int {
+		var start = 0;
+		var count = range.length;
+		while (count > 0) {
+			var step = count >> 1;
+			var mid = start + step;
+			if (!lt(val, range[mid])) { start = mid + 1; count -= step + 1; }
```
titzer, line 110: I think there is a subtle distinction between `!lt(x, y)` and `lt(y, x)`. I think this should code should use the latter. I think it should be accompanied with a comment that illustrates, like so:

```
// Assumes that {range} is sorted according to `lt` and returns the first element that is *greater* than `val`.
// e in         | elem  elem  elem  elem  elem  elem  |
// lt(e, val)   | true  true  false false false false |
// lt(val, e)   | false false false true  true  true  |
//                                  ^
```

I think all three methods could use this comment.

lib/util/Ranges.v3
```diff
@@ -78,38 +78,38 @@ component Ranges {
 	}
 	// Performs binary search and returns the index.
 	def binarySearch<T>(range: Range<T>, val: T, lt: (T, T) -> bool) -> int {
```
titzer, line 80: I think we should name these three methods `binarySearchEq`, `binarySearchLt`, and `binarySearchGt` and comment them uniformly. See below.

PR #298, opened 2024-11-18
Add F_POINTED_AT; seems to work after some minimal testing

eliotmoss:


titzer (approved):
lgtm

aeneas/src/ir/Normalization.v3
```diff
@@ -329,7 +329,8 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 				// add normalized field(s) to the vector
 				var tn = rf.typeNorm;
 				var facts = if(tn.size > 1, Fact.F_NORM, Facts.NONE);
-				if (!rf.raFacts.RF_WRITTEN) facts |= (Fact.F_VALUE | Fact.O_FOLDABLE);
+				if (!rf.raFacts.RF_WRITTEN && !rf.orig.facts.F_POINTED_AT) facts |= (Fact.F_VALUE | Fact.O_FOLDABLE);
```
titzer, line 332: Note I added this line while resolving a conflict here.

eliotmoss:
On 11/19/2024 4:33 AM, Ben L. Titzer wrote:
> I've confirmed the CI fails because the additional fact brought the size of a |Fact.set| over 32-bit, which seems to 
> have exposed some bugs in unboxing (that's why this only fails on the CI configurations that try to unbox all variants). 
> I'll try to get a repro and fix those other bugs. In the mean time, can you remove the unused |O_ASSOCIATIVE| case?

Yes - but do you mean in the PR or only locally?  Eliot

titzer:
> Yes - but do you mean in the PR or only locally? Eliot

In the PR. Since it only exposed a pre-existing bug (which I have a fix for, but not yet a smaller repro), no need to hold up this PR.

eliotmoss:
On 11/19/2024 12:02 PM, Ben L. Titzer wrote:
>     Yes - but do you mean in the PR or only locally? Eliot
> 
> In the PR. Since it only exposed a pre-existing bug (which I have a fix for, but not yet a smaller repro), no need to 
> hold up this PR.

Change developed and added ... EM

PR #299, opened 2024-11-19
Small change to the aeneas script to make it more robust

eliotmoss:
Otherwise, blocks that start with a phi do not get optimized.
An interesting question is whether there are other things in a block, besides SsaApplyOp's, that should get skipped over (or otherwise handled) in the while loop.  But this change was effective in removing junk from XXGetField's expanded in normalization from accesses to unboxed variants, at least in my test case.  Your CI can check to see whether this causes any regressions ...

### DIFF
--- a/aeneas/src/ssa/SsaOptimizer.v3
+++ b/aeneas/src/ssa/SsaOptimizer.v3
@@ -1911,6 +1911,7 @@ class SsaLoadOptimizer(context: SsaContext) {
 		pure.length = 0;
 		impure.length = 0;
 		var i = block.next;
+		if (SsaPhi.?(i)) i = i.next;
 		while (SsaApplyOp.?(i)) {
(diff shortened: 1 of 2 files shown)


titzer:
> Otherwise, blocks that start with a phi do not get optimized.
> An interesting question is whether there are other things in a block, besides SsaApplyOp's, that should get skipped over (or otherwise handled) in the while loop. But this change was effective in removing junk from XXGetField's expanded in normalization from accesses to unboxed variants, at least in my test case. Your CI can check to see whether this causes any regressions ...

Sounds good, but it looks like the PR only contains a change to the aeneas script.

eliotmoss:
Sigh -- yeah, wrong commit.  I'm new to preparing PR's.  I'll see about changing the title and content and then submitting the right PR!

PR #300, opened 2024-11-19
Fix load optimizer to skip the phi at a start of some blocks

eliotmoss:
Without this change, the load optimizer does nothing to a block that starts with a phi.
We should also consider whether there are things other than an SsaApplyOp where we should do something, or at least keep going.
(Sorry about the aeneas script change lurking in there.)


eliotmoss:
I see you've already made a changes that maximizes this one ...

PR #312, opened 2024-12-12
Don't merge -- Test to see if `finishLoop` is fixed

k-sareen:


k-sareen:
@titzer So this branch at least doesn't fail any test and doesn't go into an infinite loop. I had to disable the tests with the stable compiler for `x86-linux` because they were broken. The current code has all of my debugging changes in it. I can clean it up and submit a new PR with the fix.

The fix is just the reverse of what I was thinking about Re: forward data-flow analysis. Each `VReg` has a boolean field (`activeInFuture`) associated with it which tells us if it will have an `LsraStart` in the future. Then in `finishLoop` if we have variables that will have an `LsraStart` in the future and if the last block of the loop is the predecessor of the block after the loop, we then add the `LsraEnd`s for those variables. I wasn't sure if doing `livein.row(0).apply` was always sound in such cases, so currently I've opted to just push such variables to a `Vector` and then process them later.

 We just need to maintain the `activeInFuture` flag correctly when computing the liveness information. Of note is when we remove the interior `Lsra{Start,End}`s in `finishLoop`, we need to update this field as well as otherwise we still end up with the broken nested live ranges.

k-sareen:
See https://github.com/titzer/virgil/pull/313 for the new PR in case I'm AWOL tomorrow.

titzer:
> @titzer So this branch at least doesn't fail any test and doesn't go into an infinite loop. I had to disable the tests with the stable compiler for `x86-linux` because they were broken. The current code has all of my debugging changes in it. I can clean it up and submit a new PR with the fix. I would recommend keeping at least some of my debugging changes such as checking that an infinite loop has not been created.
> 
> The fix is just the reverse of what I was thinking about Re: forward data-flow analysis. Each `VReg` has a boolean field (`activeInFuture`) associated with it which tells us if it will have an `LsraStart` in the future. Then in `finishLoop` if we have variables that will have an `LsraStart` in the future and if the last block of the loop is the predecessor of the block after the loop, we then add the `LsraEnd`s for those variables. I wasn't sure if doing `livein.row(0).apply` was always sound in such cases, so currently I've opted to just push such variables to a `Vector` and then process them later.
> 
> We just need to maintain the `activeInFuture` flag correctly when computing the liveness information. Of note is when we remove the interior `Lsra{Start,End}`s in `finishLoop`, we need to update this field as well as otherwise we still end up with the broken nested live ranges.
> 
> I'll document all of this in comments when I clean up the PR if you find it sensible.

Thanks for running this down. I'll look at the fix in the PR and think about if there is a simpler way to do this.

PR #308, opened 2024-12-04
[bug] Compiler stuck on building x86-linux target

linxuanm:
Compiling the code in this PR halts the compiler. This was originally discovered when adding tests to the JsonParser in the standard library.

Reproduce: `cd` into `test/lib` and run `TEST_TARGETS=x86-linux ./test.bash`.

PR #313, opened 2024-12-12
[lsra]: Fix nested live ranges due to `finishLoop`

k-sareen:
In certain edge cases, `finishLoop` would not insert `LsraEnd`s for active variables resulting in nested live ranges which manifested as an infinite loop in the register allocator (see discussion here [1]).

This commit fixes this bug by keeping track of whether a variable will be made active in the future or not (i.e. if it will have a live range). If a variable will have a live range in the future, then `finishLoop` will make sure to insert `LsraEnd`s for them to avoid the above bug.

[1]: https://github.com/titzer/virgil/pull/308

### DIFF
--- a/aeneas/src/mach/LinearScanRegAlloc.v3
+++ b/aeneas/src/mach/LinearScanRegAlloc.v3
@@ -100,6 +100,10 @@ class LinearScanRegAlloc {
 	}
 	def recordLive(index: int) {
 		for (l = activeList; l != null; l = l.next) {
+			if (l == l.next) {
+				gen.mach.prog.ERROR.fail(
+					Strings.format2("VReg %d from SSAInstr %d points to itself!", l.varNum, l.ssa.uid));
+			}
 			livemap[index, l.varNum] = true;
 		}
 	}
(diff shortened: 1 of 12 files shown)


k-sareen:
@titzer Feel free to edit this PR directly if you think it's a good baseline. The hardest part for me was to reason about `activeInFuture` when removing the interior `Lsra{Start,End}`s in `finishLoop`: https://github.com/titzer/virgil/pull/313/files#diff-733fa4b1a9c66210d8c6e3b0b156b605203d2a9b0feb6aa1fec7a96141274fbcR555-R566

I'm not entirely sure what I have is correct because surely this is not the only test/program which has variables whose live range ends inside a loop and then are used in the future as well. I worry this change has artificially increased the live range for certain cases, but I don't know how I would have checked that.

titzer:
I noticed that splitting critical edges before codegen fixes all of these issues. I am planning on doing a stable rev of the compiler and have basically been whacking bugs and making things more stable. Thus a simpler and less invasive change is attractive here.

k-sareen:
Okay sure. I think LSRA still needs debug assertions to prevent this from slipping through

PR #315, opened 2024-12-14
Add support for assembler tracing

eliotmoss:
Adds assembler tracing with a build time flag to enable and a run-time flag to turn on/off


eliotmoss:
Are you prepared to accept this PR yet?  If there is more work needed, just let me know.

PR #321, opened 2025-01-05
[lib] Added `HashMap.remove` to `Map.v3`

linxuanm:
This PR adds support for removing a key-value pair from hash map along with a basic test for `lib/util/Map.v3`.

This would be useful for the Virgil language server that I'm working on, which needs a LRU implementation (with hashmap) and I thought it would be useful to be in the standard util library.

### DIFF
--- a/lib/util/Map.v3
+++ b/lib/util/Map.v3
@@ -65,9 +65,6 @@ class HashMap<K, V> extends PartialMap<K, V> {
 			// missed the cache, table not yet allocated, insert both
 			table = Array.new(16); // TUNABLE: initial HashMap table size
 			if (c != null) insert(c);
-			
-			var v = hash(key); // NPE here
-
 			insert(cache = Bucket.new(key, val, null));
 			return;
 		}
--- a/lib/util/Map.v3
+++ b/lib/util/Map.v3
@@ -112,7 +112,47 @@ class HashMap<K, V> extends PartialMap<K, V> {
 	}
 	// Remove {key} from this map. Return true if {key} existed.
 	def remove(key: K) -> bool {
-		return false;
+		var c = cache;
+		if (c == null) return false; // empty map
+		var found = false;
+
+		// search table
+		if (table != null) {
+			for (i < table.length) {
+				if (table[i] == null) continue;
+				if (table[i].key == key || equals(table[i].key, key)) {
+					table[i] = table[i].next;
+					found = true;
+					break;
+				}
+				for (b = table[i]; b.next != null; b = b.next) {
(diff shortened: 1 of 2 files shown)


lib/util/Map.v3
```diff
@@ -110,6 +110,50 @@ class HashMap<K, V> extends PartialMap<K, V> {
...
+						b.next = b.next.next;
+						found = true;
+						break;
+					}
+				}
+				if (found) break;
+			}
+		}
+
+		// if cache hit, clear cache and set it to any bucket
```
titzer, line 139: I think it's better to just clear the cache and let the next access refill it.

lib/util/Map.v3
```diff
@@ -110,6 +110,50 @@ class HashMap<K, V> extends PartialMap<K, V> {
...
+	// Remove {key} from this map. Return true if {key} existed.
+	def remove(key: K) -> bool {
+		var c = cache;
+		if (c == null) return false; // empty map
+		var found = false;
+
+		// search table
+		if (table != null) {
+			for (i < table.length) {
+				if (table[i] == null) continue;
```
titzer, line 122: I think you can pull the `b` variable up to here; it's the head of the bucket list.

test/lib/MapTest.v3
```diff
@@ -0,0 +1,125 @@
...
+def test_get(t: LibTest) {
+	var c = Counter.new();
+	var i_s = int_map<string>();
+	var s_s = Strings.newMap<string>();
+	for (i < 100) {
+		var s = int2str(i);
+		i_s[i] = s;
+		s_s[s] = s;
+	}
+	for (i < 100) {
```
titzer, line 43: You might want to test the strings by identity, e.g. by keeping an array of pairs here.

titzer (commented):
lgtm modulo minor comments, and fixing tests.

lib/util/Map.v3
```diff
@@ -110,6 +107,39 @@ class HashMap<K, V> extends PartialMap<K, V> {
...
+		var c = cache;
+		var found = false;
+		// if cache hit, clear cache
+		if (c != null && (c.key == key || equals(c.key, key))) {
+			found = true;
+			cache = null;
+		}
+		// search table
+		if (table != null) {
+			for (i < table.length) {
```
titzer, line 121: Sorry I didn't notice this before. This should do a hashed search; no need to search the entire table, only the bucket chain in the right hash slot.

linxuanm:
Ah sorry I must've missed it. This should be fixed now.

titzer (approved):

PR #326, opened 2025-01-07
Add split/strip utilities for Strings

ejrgilbert:


lib/util/Arrays.v3
```diff
@@ -19,6 +19,17 @@ component Arrays {
...
+	def strEqual(x: Array<string>, y: Array<string>) -> bool {
+		return equal_with(x, y, Strings.equal);
+	}
+	// Check equality of {x} and {y} with custom equality function.
+	def equal_with<A>(x: Array<A>, y: Array<A>, eq: (A, A) -> bool) -> bool {
+		if (x == y) return true;
+		if (x.length != y.length) return false;
+		for (i < x.length) if (!eq(x[i], y[i])) return false;
+		return true;
+	}
```
ejrgilbert, line 32: It'd be better to refactor `equal` to use this `equal_with` passing in a closure...but I don't know if there's an inline way to define a closure? e.g.

```
def equal<A>(x: Array<A>, y: Array<A>) -> bool {
  return equal_with(x, y, |a, b| -> bool {
    a == b;
  });
}
```

ejrgilbert: I guess that'd just be the function that handles the `==` operation...I'll check to see if that's available somewhere to pass.

ejrgilbert: Oh daaaang...that's niiiiice

`return equal_with(x, y, A.==);`

lib/util/Strings.v3
```diff
@@ -43,6 +50,77 @@ component Strings {
...
 	}
+	// Check if the string is empty.
+	def isEmpty(str: string) -> bool {
+		return str.length <= 0;
+	}
+	// Splits a string into an array of strings.
+	// You can specify the separator, default separator is any whitespace.
+	// separator: Specifies the separator to use when splitting the string. By default whitespace is the separator. Pass empty string if using default. 
+	// maxsplit: Specifies how many splits to do. By default, split at "all occurrences." Pass -1 if using default.
+	def split(str: string, separator: string, maxsplit: int) -> Array<string> {
```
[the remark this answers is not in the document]
ejrgilbert: Responded to this in slack so it'd be easier to have a back and forth.

lib/util/Arrays.v3
```diff
@@ -14,9 +14,17 @@ component Arrays {
 	}
 	// Compare elements of {x} and {y} for equality.
 	def equal<A>(x: Array<A>, y: Array<A>) -> bool {
+		return equal_with(x, y, A.==);
+	}
+	// Compare string elements of {x} and {y} for equality.
+	def strEqual(x: Array<string>, y: Array<string>) -> bool {
```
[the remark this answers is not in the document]
ejrgilbert: I feel like it's more readable to have the `equal_with` method. Are you okay with keeping that method to be able to call this in the caller:
`Arrays.equal_with(x, y, Strings.equal)`

I can refactor both `equal_with` and `equal` to be using `allTrue` in the background.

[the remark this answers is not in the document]
titzer: I think I'd prefer to keep the `allTrue` and not introduce a wrapper around it. Also, `Arrays.equal` is called a lot, so having with the inline use of `==` will have better performance.

lib/util/Strings.v3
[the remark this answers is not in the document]
ejrgilbert: Splitting on whitespace has a bit of a nuance to it that wouldn’t be the same logic as splitting on characters:
If we were to split this string on whitespace " a a        a", you’d expect: `["a", "a", "a"]`
If we were to split this string using the whitespace char predicate, you’d actually get: `["", "a", "", "", …, "a"]`
(that’s why the whitespace handling code uses the stripLeading  function, it keeps you from getting all of those empty strings.)

So this needs to be a special case. I will split out to 3 functions:
```
splitOnWhitespace : handles the whitespace special case
splitOnChar : pass char predicate
split : pass string separator
```

lib/util/Strings.v3
```diff
@@ -43,6 +34,118 @@ component Strings {
...
+		}
+		if (separator.length == 1) {
+			// split on single character
+			return splitOnChar(str, byte.==(separator[0], _), maxsplit);
+		}
+
+		return split_internal(str, maxsplit, split_cond(str, separator, _), separator.length);
+	}
+	private def split_cond(str: string, separator: string, idx: int) -> bool {
+		return startsWith(Arrays.range(str, idx, str.length), separator);
```
titzer, line 106: That looks like it makes a copy of the string from the current index until the end, so that'll be expensive. Maybe just inline a byte-by-byte comparison here to avoid it?

titzer (approved):
lgtm if minor performance issue is addressed.

lib/util/Strings.v3
titzer: Yep, sgtm.

titzer (approved):
lgtm, let's light this rocket

PR #329, opened 2025-01-08
Add another Ranges util for lastIndexOf

ejrgilbert:
I realized that this would be useful when working more on the linking PR comments.

### DIFF
--- a/lib/util/Ranges.v3
+++ b/lib/util/Ranges.v3
@@ -94,6 +94,21 @@ component Ranges {
 	def endsWithFrom<T>(range: Range<T>, start: int, end: Range<T>) -> bool {
 		return (range.length - start) == end.length && endsWith(range, end);
 	}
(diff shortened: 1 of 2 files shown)


lib/util/Ranges.v3
```diff
@@ -94,6 +94,21 @@ component Ranges {
 	def endsWithFrom<T>(range: Range<T>, start: int, end: Range<T>) -> bool {
 		return (range.length - start) == end.length && endsWith(range, end);
 	}
+	// Find the last index of some value {v} in a {range}.
+	// Return -1 if not found.
+	def lastIndexOf<T>(range: Range<T>, v: T) -> int {
+		var i = range.length - 1;
+		var found = false;
+		while (i >= 0) {
+			if (v == range[i]) {
```
titzer, line 103: Could `return i` here and `-1` outside.

titzer (approved):

PR #330, opened 2025-01-11
[lib] Fixed assymetry between `Ranges.startsWith` and `Ranges.endsWith`

linxuanm:
It seems like `startsWith` and `endsWith` currently exhibits different behaviors in that a range can be a prefix of itself in `startsWith`, but cannot be a suffix of itself in `endsWith`. This PR changes `endsWith` to match with `startsWith`.


titzer (approved):
Can you add a unit test so the expected behavior is also tested?

titzer (approved):

PR #342, opened 2025-02-05
Refining and cleaning up Pointer.atField for unboxed fields, with more tests, etc.

eliotmoss:


aeneas/src/ir/Reachability.v3
```diff
@@ -415,18 +415,42 @@ class ReachabilityAnalyzer(compilation: Compilation) {
 			getEquality(if(rf.fieldType == null, f.fieldType, rf.fieldType), null);
 		}
 	}
-	def getAndSetField(rf: RaField) {
+	def pointedAtTypeLegal(t: Type) -> bool {
+		if (PrimType.?(t) && !VoidType.?(t)) return true;
```
titzer, line 419: A nice opportunity to use a match with a typecase :-)

aeneas/src/ir/Reachability.v3
```diff
@@ -727,6 +761,9 @@ class RaField(receiver: Type, orig: IrField, fieldType: Type) extends RaItem {
 		var rc = ra.getClass(receiver);
 		return normOf(rc.liveFields);
 	}
+	def getClosedType() -> Type {
```
titzer, line 764: Probably should have written this helper ages ago.

test/pointer/seman/Pointer_atFieldUnboxedBad20.v3
```diff
@@ -0,0 +1,13 @@
+//@seman=TypeError @ 11:17
+// tests possible issues related to unboxed field and parameterized types
+
+component Comp {
+	var f: void;
+}
+
+def main(n: int) -> int {
```
titzer, line 8: Seman tests don't need to have a `main`, because they don't execute. We can simplify this down to just:
```
component C {
    var f: void;
    var p = Pointer.atField(f);
}
```

titzer: In general, the fewer lines of code needed to get to a semantic error, the better. Helps a lot in debugging and makes tests run faster.

PR #352, opened 2025-02-13
[lib/x86-64-linux] Update statx layout + add O_DIRECTORY

vouillon:
Provide access to the subsecond part of timestamps


titzer (approved):
lgtm, thanks!

PR #355, opened 2025-02-24
WASM GC backend

eliotmoss:
Fix batch of changes being submitted for comment.  Work ends essentially just after the point where mtables are processed to obtain function table information.


aeneas/src/wasm/WasmCommon.v3
```diff
@@ -0,0 +1,230 @@
+// Copyright 2016 Google Inc. All rights reserved.
```
titzer, line 1: Copyright notice.

aeneas/src/core/Program.v3
```diff
@@ -22,6 +22,7 @@ class Program {
 	var mainRootIndex = -1;
 	var explicitName: string;
 	var tprog: TargetProgram;
+	var ra: ReachabilityAnalyzer;	// kept upon request
```
titzer, line 25: I'd prefer not keeping the whole analyzer alive; instead we should think about the result of the analysis that is important and only preserve that.

aeneas/src/ir/Ir.v3
```diff
@@ -40,6 +41,7 @@ class IrClass extends IrItem {
 		}
 		return m;
 	}
+	def accept<E, R>(v: IrVisitor<E, R>, env: E) -> R { return v.visitIrClass(this, env); }
```
titzer, line 44: I think I'd prefer not introducing visitors for this; a match on the type is locally more compact and usually keeps the logic in one place.

aeneas/src/main/CLOptions.v3
```diff
@@ -150,6 +150,8 @@ component CLOptions {
 		"Generate dwarf line debug information only in the compiled binary.");
 	def RA			= sharedOpt.newBoolOption("ra", false,
 		"Enable reachability analysis and normalization before running the program.");
+	def SAVE_RA		= sharedOpt.newBoolOption("save-ra", false,
```
titzer, line 153: If this is configured by the backend, it probably doesn't need a CL options, probably just a compiler option.

aeneas/src/wasm/WasmType.v3
```diff
@@ -48,32 +48,174 @@ type WasmType {
...
+		}
+		w.putb(if(mut, WasmMutabilityCode.Mutable.code, WasmMutabilityCode.Constant.code));
+	}
+	def hash() -> int {
+		match (this) {
+			RefNull(ht) => return this.tag + int.view(ht.index);
+			_ => return this.tag;
+		}
+	}
+	def equals(other: WasmValueType) -> bool {
```
titzer, line 65: IIUC there shouldn't be a need for an equals method; variants get a `==` that is a deep structural comparison. As long as every case is OK with that (looking at the `RefNull` case).

titzer (commented):
First comments.

aeneas/src/wasm/WasmCommon.v3
eliotmoss: Changed to add: Eliot Moss 2024-2025.  If I should be using something different, let me know ...

aeneas/src/core/Program.v3
eliotmoss: Turns out I wasn't using that.  Must be a relic from some direction I did not pursue.  Removed.

aeneas/src/main/CLOptions.v3
eliotmoss: Likewise not used, so removed.

aeneas/src/wasm/WasmType.v3
eliotmoss: The `RefNull` case requires equality on `WasmHeapType`s, and they are classes, not ADTs.  So I left this for now.

eliotmoss:
This was the working working area that led to a now merged (different) PR for the Wasm Gc back end.  So I am closing this one.

PR #368, opened 2025-03-01
Improve aeneas scripts

eliotmoss:
This change make all relevant bash scripts include a top-level funcs.bash script.  At present, that script defines bash functions for cd and pwd, which avoid vulnerabilities if people like me redefine those commands and/or use CDPATH.


eliotmoss:
The work this was aiming to do is in a new branch, with a PR that is passing tests.

PR #373, opened 2025-03-05
Improve scripts (pull request)

eliotmoss:
Chanfge scripts to protect against user redefinition of cd, etc.


eliotmoss:
I struggled to find why the wasm builds were failing, but finally got it.  I realize now that there are more scripts out there that could be reworked, but I think I'll stop for now since I think most of them are ci scripts so usually run in a more controlled environment.

PR #374, opened 2025-03-06
Add x86-64 syscall numbers to LinuxConst

MSMazaya:


titzer (commented):
I think it makes sense to do a batch addition of all the system call numbers, wdyt?

titzer (approved):
lgtm; rubber-stamped

PR #380, opened 2025-03-18
fix duplicate DefCase print in VstPrinter

mkhan45:
I'm not sure if this is the correct fix, since I don't really understand what we expect `defcase` to be. If it's meant to be for `match else`, the bug is somewhere upstream. If it's supposed to point to a wildcard in `stmt.cases` this makes sense though.

```
            (MatchStmt
                (VarExpr[Local] "kind": FieldKind)
                    (VarExpr[None] "U8")
                    (...)
                    ...
                    (DefCase)
                    (ReturnStmt
                        (AppExpr "invalid": void
                            (VarExpr[ComponentMethod] "invalid": void -> void)
                            (VarExpr[Local] "invalid": CanonicalDefs)))
                    (DefCase) /* ! Duplicate ! */
                    (ReturnStmt
                        (AppExpr "invalid": void
                            (VarExpr[ComponentMethod] "invalid": void -> void)
                            (VarExpr[Local] "invalid": CanonicalDefs)))))

```

### DIFF
--- a/aeneas/src/vst/VstPrinter.v3
+++ b/aeneas/src/vst/VstPrinter.v3
@@ -268,7 +268,6 @@ class VstPrinter extends VstVisitor<int, void> {
 		p.enter("MatchStmt", indent);
 		stmt.expr.accept(this, indent + 1);
 		Lists.apply(stmt.cases.list, printCases(_, indent + 2));
-		if (stmt.defcase != null) printCases(stmt.defcase, indent + 2);
 		p.exit();
 	}
 	def visitEmpty(stmt: EmptyStmt, indent: int) {


aeneas/src/vst/VstPrinter.v3
```diff
@@ -268,7 +268,6 @@ class VstPrinter extends VstVisitor<int, void> {
 		p.enter("MatchStmt", indent);
 		stmt.expr.accept(this, indent + 1);
 		Lists.apply(stmt.cases.list, printCases(_, indent + 2));
-		if (stmt.defcase != null) printCases(stmt.defcase, indent + 2);
```
titzer, line 271: I think `defcase` won't be in the list in the case of `match () { } ... else { }`. So I think the right fix is to check it's been printed from above. So you should probably make the above `apply` into an explicit loop and check if the `defcase` was already printed.

aeneas/src/vst/VstPrinter.v3
```diff
@@ -267,8 +267,14 @@ class VstPrinter extends VstVisitor<int, void> {
...
 		stmt.expr.accept(this, indent + 1);
-		Lists.apply(stmt.cases.list, printCases(_, indent + 2));
-		if (stmt.defcase != null) printCases(stmt.defcase, indent + 2);
+
+                var printedDefcase = false;
+                for (cur_case = stmt.cases.list; cur_case != null; cur_case = cur_case.tail) {
+                    printCases(cur_case.head, indent + 2);
+                    printedDefcase |= (cur_case.head.patterns == null);
+                }
+		if (stmt.defcase != null && !printedDefcase) printCases(stmt.defcase, indent + 2);
```
mkhan45, line 276: This seems to work

aeneas/src/vst/VstPrinter.v3
```diff
@@ -267,8 +267,14 @@ class VstPrinter extends VstVisitor<int, void> {
 	def visitMatch(stmt: MatchStmt, indent: int) {
 		p.enter("MatchStmt", indent);
 		stmt.expr.accept(this, indent + 1);
-		Lists.apply(stmt.cases.list, printCases(_, indent + 2));
-		if (stmt.defcase != null) printCases(stmt.defcase, indent + 2);
+
+                var printedDefcase = false;
+                for (cur_case = stmt.cases.list; cur_case != null; cur_case = cur_case.tail) {
+                    printCases(cur_case.head, indent + 2);
+                    printedDefcase |= (cur_case.head.patterns == null);
```
titzer, line 274: You could also check `cur_case == stmt.defcase`.

titzer (approved):
lgtm

PR #381, opened 2025-03-18
Add s-expression parser to lib/util

mkhan45:
I don't remember if you thought this should be added to util, but:

```
    def parser = SexpParser.new(filename, file_bytes);
    def res = parser.readSexp();
    match (res) {
        Success(res) => {
            ...
        }
        err => {
            ...
        }
    }
```

### DIFF
--- a/lib/util/Sexpr.v3
+++ b/lib/util/Sexpr.v3
@@ -1,0 +1,75 @@
+type Sexp {
+    case Atom(data: string);
+    case List(elts: Range<Sexp>); // usually Cons
+
+    def buildString(buf: StringBuilder) {
+        match (this) {
+            Atom(data) => buf.puts(data);
+            List(elts) => {
+                buf.putc('(');
+                for (i = 0; i < elts.length - 1; i += 1) {
+                    elts[i].buildString(buf);
+                    buf.sp();
+                }
+                elts[elts.length - 1].buildString(buf);
+                buf.putc(')');
+            }
+        }
+    }
+}
+
+type ParseResult<T> {
+    case Success(res: T);
+    case Expected(str: string);
+    case EmptySexp;
+}
+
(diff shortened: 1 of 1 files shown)


lib/util/Sexpr.v3
```diff
@@ -0,0 +1,75 @@
...
+    }
+}
+
+type ParseResult<T> {
+    case Success(res: T);
+    case Expected(str: string);
+    case EmptySexp;
+}
+
+class SexpParser extends TextReader {
```
titzer, line 27: Because this is a parser that might be fed very large files, having position information for errors is really useful. Given that I think it makes sense that this parser take either an error generator object or has a mutable field for a callback on errors, that includes location information.

Then the parser can return a default value for the ADT (atom of an empty string) if it fails.

lib/util/Sexpr.v3
```diff
@@ -0,0 +1,75 @@
+type Sexp {
```
titzer, line 1: Maybe `SExpr` as a name?

lib/util/Sexpr.v3
mkhan45: Does it make sense to just exploit the underlying TextReader's error fields? 

```
    def readAtom() -> SExpr {
        skipWhitespace();
        def atom_end = star_rel(1, atomChar);
        if (atom_end == -1) {
	    fail("Expected atom");
	    return SExpr.Atom("");
	}

        def tok = readToken(atom_end - pos);
        skipWhitespace();
        return SExpr.Atom(tok.image);
    }
```

titzer (approved):

mkhan45:
@titzer, I didn't update the error handling yet: https://github.com/titzer/virgil/pull/381#discussion_r2017684906

PR #382, opened 2025-03-27
Improvements to virgil code, hopefully cleaner of other junk that previous attempt

eliotmoss:


eliotmoss:
Clearly there is something I am not "getting" about how to do this.  Will consult with Ben in the morning.

PR #386, opened 2025-04-02
Var rep hints / VstPrinter update

mkhan45:
For `var_rephint.v3`, it prints:

```
        (BlockStmt
            (LocalStmt
                ("x": int (repHints #somehint #another)
                    (Literal "12": int)))
            (ReturnStmt
                (VarExpr[Local] "x": int)))
```

I'm going to make all the printing more structured like so:
```
        (BlockStmt
            (LocalStmt
                ("x" (type int) (repHints #somehint #another)
                    (Literal "12" (type int))))
            (ReturnStmt
                (VarExpr[Local] "x" (type int))))
```

I'll also fix the nesting issue we saw with MatchStmt


mkhan45:
im not sure how the test scripts work but maybe there should be some golden/characterization tests for the printer

titzer (approved):

PR #403, opened 2025-05-02
Add sysinfo layout and wasm memory offset CLOptions

MSMazaya:


aeneas/src/main/CLOptions.v3
```diff
@@ -190,6 +190,8 @@ component CLOptions {
 		"Export the Wasm memory from generated modules.");
 	def WASM_PAGE_SIZE	= wasmOpt.newSizeOption("wasm.page-size", 65536,
 		"Specify the Wasm memory page size (must be a power of 2).");
+	def WASM_MEMORY_OFFSET	= wasmOpt.newSizeOption("wasm.mem-offset", 0,
```
titzer, line 193: Can we call this "reserved-data-size" (similar to reserved code size) in terms of bytes and then in the Wasm target, round this up to the next page size?

titzer (approved):
lgtm with suggestion

aeneas/src/main/CLOptions.v3
MSMazaya: Renamed and rounded to the next page size.

aeneas/src/wasm/WasmTarget.v3
```diff
@@ -261,7 +261,8 @@ class WasmTarget extends Target {
 		if (pageSize != 65536) flags |= 1 << 3;		// custom page size
 		out.putb(flags);				// flags = contains maximum
 		out.put_uleb32(pages);				// initial memory size
-		out.put_uleb32(pages);				// maximum memory size
+		var reserved_data_size = ((CLOptions.WASM_RESERVED_DATA_SIZE.get() + pageSize - 1u) / pageSize) * pageSize;
```
titzer, line 264: There is `Alignment.alignUp` you can use here.

aeneas/src/main/CLOptions.v3
```diff
@@ -190,6 +190,8 @@ component CLOptions {
 		"Export the Wasm memory from generated modules.");
 	def WASM_PAGE_SIZE	= wasmOpt.newSizeOption("wasm.page-size", 65536,
 		"Specify the Wasm memory page size (must be a power of 2).");
+	def WASM_RESERVED_DATA_SIZE = wasmOpt.newSizeOption("wasm.reserved-data-size", 0,
```
titzer, line 193: Just make it `reserved-data-size` and drop the `wasm.`; we might make use of this option on other platforms too.

titzer (approved):
lgtm

PR #449, opened 2025-06-23
Add vctags coverage for enums, layouts, and packings, and for paramet…

eliotmoss:
…ers to classes, enums, and layouts


titzer (approved):
lgtm, thanks!

PR #450, opened 2025-06-26
[lib/util] Add PrefixTrie implementation

titzer:


titzer:
Going to hold off on merging this functionality, as it is only used in one Wizard monitor.

PR #453, opened 2025-07-03
WIP derive to_sexpr method for types

mkhan45:
I want this mostly for easier print debugging

```scala
class TestClass {
    def x: int;
    var y: string;
    var z: TestClass;
}

class Point(x: int, y: int) {}

def main() {
    def x = TestClass.new();
    System.puts(x.to_sexpr().toString());
    System.ln();
    x.z = TestClass.new();
    x.z.y = "hello";
    System.puts(x.to_sexpr().toString());
    System.ln();

    def p = Point.new(5, 10);
    System.puts(p.to_sexpr().toString());
    System.ln();

    def ls = Lists.cons3(5, 10, 15);
    System.puts(ls.to_sexpr().toString());
    System.ln();
}
```
```
(TestClass (x 0) (y Array<byte>) (z null))
(TestClass (x 0) (y Array<byte>) (z (TestClass (x 0) (y Array<byte>) (z null))))
(Point (x 5) (y 10))
(List (head T) (tail (List (head T) (tail (List (head T) (tail null))))))
```

The main issues are:
- it doesn't support generics, because it doesn't know if a typevar will have `.to_sexpr()`. I guess this could be fixed by generating a `has_sexpr() -> bool` or preferably by generating a different `to_sexpr` per typevar instantiation
- it doesn't print strings, because I don't think it can distinguish between strings and byte arrays

### DIFF
--- a/aeneas/src/vst/Verifier.v3
+++ b/aeneas/src/vst/Verifier.v3
@@ -504,6 +504,32 @@ class VstCompoundVerifier {
 			else decl.memberMap[mdecl.token.image] = mdecl;
 			mdecl.receiver = decl;
 		}
+		if (!decl.memberMap.has("to_sexp") && !VstLayout.?(decl)) {
+			var name = Token.new(decl.token.fileName, "to_sexpr", decl.token.beginLine, decl.token.beginColumn);
+			var superclause = decl.superclause;
+			var sexpr_token = decl.token.copy("SExpr");
+			var sexpr_type_con = decl.typeEnv.lookup("SExpr");
+			if (sexpr_type_con != null) {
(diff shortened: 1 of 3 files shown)


mkhan45:
I guess this would go well with #429 but I'd like to have it before that

mkhan45:
The code this generates looks something like:

```
def to_sexpr() -> SExpr {
    return SExpr.List(List.new(
        SExpr.Atom("$TYPE.NAME"), 
        Lists.fromArray([SExpr.cons2(SExpr.Atom("$FIELD.NAME"), $FIELD.to_sexpr()) for $FIELD in $FIELDS]),
    ));
}
```
but it also checks if the field is null if it is a class

Issue #74, opened 2022-07-23
lib/util: Implement %f to print floating point numbers


titzer:
Since adding floating point back in 2020, the Virgil utility libraries have not yet added support for printing/rendering floating point numbers (in decimal). In C, the `printf` format string supports `%f` for specifying an argument is a floating point number. The analogous place to add code to handle this is in [StringBuilder](https://github.com/titzer/virgil/blob/master/lib/util/StringBuilder.v3), which can deal with floating point strings.

Slight differences with C printf and Virgil `StringBuffer`:

- `StringBuilder` uses `%d` (for *d*ecimal) for specifying decimal output of integer values. I think it'd be nice to just extend 
- `StringBuilder` doesn't yet support specifying the width (in characters) of the output item, nor left or right-justifying
- Performance of printing floats isn't *highly* important; as long as the routine doesn't allocate memory, it should be fine to do the naive algorithm

I could use some help on this, and it might make a good starter project for someone new wanting to kick the tires with Virgil and contribute.

srackham:
I've got an implementation working but need some input:

1. printf's `%f` renders the fractional part rounded to 6 decimal places with trailing zeros. I think a more readable default format would be rendering to 6 places but truncating trailing zeros after the first decimal place? For example: this `1.0` and not this `1.000000`.

2. Floating point integer parts become unsafe after these values:

        def MAX_SAFE_DOUBLE = 9007199254740991.0;	// 2^53 - 1
        def MAX_SAFE_FLOAT = 16777215.0f;		// 2^24 - 1

    See, for example, https://stackoverflow.com/a/1848762/1136455

    The question becomes how to report unsafe conversions? My current solution is to render the string `!UnsafeConversion` instead of the number.

titzer:
Great that you are working in this!

I haven't quite thought through all of the issues here, so happy to have your input.

1. I was thinking that `%f` would format floating point values according to the default Java rules which generate the shortest literal string that when parsed and rounded according to the existing `string -> float/double` rules will produce the exact same floating point bits as the input. Thus (modulo `NaN`s), `parse(print(x)) == x`.

[...]

PR #455, opened 2025-07-10
WIP: Basic algorithm for printing floats with %f

JohnnyDingYQ:
[Issue #74](#74)
This is as WIP pull request that aims to enable string formatting of strings. I implemented the "basic conversion algorithm" outlined in the [Ryu paper](https://dl.acm.org/doi/pdf/10.1145/3192366.3192369). A future port to Ryu, the improved algorithm, can be planned in the future once the existing code is stable

The major additions are:
- Created [lib/util/Floats.v3](lib/util/Floats.v3) for core floating rendering code
- Modified [test/lib/StringBuilderTest.v3](test/lib/StringBuilderTest.v3) for %f tests

### DIFF
--- a/lib/util/Floats.v3
+++ b/lib/util/Floats.v3
@@ -1,0 +1,220 @@
+// Credit to the ryu algorithm: https://dl.acm.org/doi/pdf/10.1145/3192366.3192369
+
+component Floats {
+	private def FLOAT_EXP_WIDTH: byte = 8;
+	private def FLOAT_SIGN_WIDTH: byte = 1;
+	private def FLOAT_MANTISSA_BITS: byte = 23;
+	private def FLOAT_BIAS: byte = 127;
+
+	private def DEBUG: bool = true;
+
+    def renderFloat(val: float, a: Array<byte>, pos: int) -> int {
+		var length: int = 0;
+        if (u32.view(val) >> byte.!(FLOAT_MANTISSA_BITS + FLOAT_EXP_WIDTH) == 1u) {
+			a[pos++] = '-';
+			return 1 + renderPosFloat(val, a, pos);
+		}
+		return renderPosFloat(val, a, pos);
+    }
+	
+
+	def renderPosFloat(val: float, a: Array<byte>, pos: int) -> int {
+		var bits = u32.view(val);
+		var exponent: u32 = (bits >> FLOAT_MANTISSA_BITS) & ((1u << FLOAT_EXP_WIDTH) - 1);
+		var mantissa: u32 = bits & ((1u << FLOAT_MANTISSA_BITS) - 1);
+
+		if (DEBUG)
+			System.puts(Strings.format2("Exponent: %d | Mantissa: %d \n", exponent, mantissa));
+
+		if (exponent == u32.view(0b11111111) || (exponent == 0 && mantissa == 0)) {
+			return renderFloatSpecial(exponent, mantissa, a, pos);
+		}
+
+		return renderPosFloatNaive(exponent, mantissa, a, pos);
+	}
+
+
+	def renderFloatSpecial(exponent: u32, mantissa: u32, a: Array<byte>, pos: int) -> int {
+		if (mantissa != 0) {
+			a[pos++] = 'N';
+			a[pos++] = 'a';
+			a[pos] = 'N';
+			return 3;
+		}
+		if (exponent != 0) {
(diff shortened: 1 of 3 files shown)


lib/util/Floats.v3
```diff
@@ -0,0 +1,377 @@
+// Credit to the ryu algorithm: https://dl.acm.org/doi/pdf/10.1145/3192366.3192369
```
titzer: Please add a copyright header assigning the copyright to "Virgil authors".

lib/util/StringBuilder.v3
```diff
@@ -351,6 +351,11 @@ class StringBuilder {
 		return (-1, offset);
 	}
 
+	def putf(f: float) -> this {
```
titzer: This is fine for now, but we'll probably want to also integrate float rendering into the `%d` case so that it works automatically for passing floats and doubles in more format strings.

lib/util/StringBuilder.v3
```diff
@@ -351,6 +351,11 @@ class StringBuilder {
...
+		acquire(16);
```
titzer: IIUC the worst case might be longer than 16?

JohnnyDingYQ: Yes, my new guess is 41, which is len(float_max) + sign + dot = 39 + 1 + 1 = 41. I can validate this by adding more tests.

lib/util/Floats.v3
```diff
@@ -0,0 +1,665 @@
+// Copyright 2024 Virgil authors. All rights reserved.
+// See LICENSE for details of Apache 2.0 license.
+
+// ====================== Start of Ryu license ========================
```
titzer: Since both Virgil and Ulf used the Apache 2.0 license, you can combine these notices, e.g.

```
// Copyright 2025 Virgil authors.
// Copyright 2018 Ulf Adams (see below).

...

// Based on code by Ulf Adams (further reference)
```

lib/util/Floats.v3
```diff
@@ -0,0 +1,665 @@
...
+
+
+	def renderFloatSpecial(exponent: u32, mantissa: u32, a: Array<byte>, pos: int) -> int {
+		if (mantissa != 0) {
+			a[pos++] = 'N';
+			a[pos++] = 'a';
+			a[pos] = 'N';
+			return 3;
+		}
+		if (exponent != 0) {
```
titzer, line 62: You could use `Arrays.copyInto` for these sequences.

Also side note: I was planning on refactoring both `Ints` and `Longs` to use `Range<byte>` but until then, it's good that you're consistent with them.

lib/util/StringBuilder.v3
titzer: Thanks for adding the case, though I've reconsidered and we should add that one line in a following PR to avoid including the code by default (DCE won't necessarily be smart enough to remove it just yet).

lib/util/StringBuilder.v3
```diff
@@ -41,13 +41,14 @@ class StringBuilder {
...
 	}
-	// Append an integer (of any width) as decimal.
+	// Append an integer (of any width) or float as decimal.
 	def putd<T>(i: T) -> this {
 		match (i) {
 			x: u32 => putd_u32(x);
 			x: i32 => putd_i32(x);
 			x: u64 => putd_u64(x);
 			x: i64 => putd_i64(x);
+			x: float => putf(x);
```
titzer: A line for a future PR.

titzer (commented):
Looks good with some minor comments.

Thanks again!

JohnnyDingYQ:
Hi, I have moved the tests into a new file `FloatsTest.v3` and completed the function that would permit the `%.Nd` feature. The relevant tests are presents and passing, so I think this PR is good to go. I reverted changes in `StringBuilder.v3` and `StringBuilderTest.v3` and kept all modifications in `Floats.v3` and `FloatsTest.v3` (new files).

titzer:
> Hi, I have moved the tests into a new file `FloatsTest.v3` and completed the function that would permit the `%.Nd` feature. The relevant tests are presents and passing, so I think this PR is good to go. I reverted changes in `StringBuilder.v3` and `StringBuilderTest.v3` and kept all modifications in `Floats.v3` and `FloatsTest.v3` (new files).

Thanks!

PR #457, opened 2025-07-12
Tests aimed at wasm (copied from wasm-gc) including 4 that fail (in f…

eliotmoss:
Four tests that fail for wasm are in test/wasm/fail.

### DIFF
--- a/test/wasm/arrays00.v3
+++ b/test/wasm/arrays00.v3
@@ -1,0 +1,15 @@
+//@execute 0=0; 4=6;10=45
+
+def main(len: int) -> int {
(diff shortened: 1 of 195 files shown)


eliotmoss:
On 7/12/2025 9:22 AM, Ben L. Titzer wrote:
> *titzer* left a comment (titzer/virgil#457) <https://github.com/titzer/virgil/pull/457#issuecomment-3065403373>
> 
> I'll merge this rubber-stamped. Many of these tests can probably move to more general directories, but we can do that 
> incrementally as needed.

Agreed.  I meant it as kind of a temporary holder directory.

PR #461, opened 2025-07-29
explicit simple fun body ret type

mkhan45:
Adds support for explicitly typing the return in simple fn bodies:
```scala
def pop_i32() => stack.pop().unbox_i32();
def pop_i32() -> TypeVar.I32 => stack.pop().unbox_i32();
```
It's a bit ugly but I'd like it for CBD stuff where there are a lot of simple functions where the types can be a bit confusing without annotations.

It can also help when inference is too precise, this is a contrived example but I ran into it before somewhere I can't remember:
```scala
type SomeADT {
	case X;
	case Y;
}

def foo() => SomeADT.X; // foo inferred as void -> SomeADT.X
// def foo() -> SomeADT => SomeADT.X; // would typecheck properly

def main() {
	var a = foo();
	System.puts(if(a == SomeADT.X, "x", "y"));
	System.ln();
	a = SomeADT.Y; // type error
	System.puts(if(a == SomeADT.Y, "y", "x"));
	System.ln();
}
```

It might look cleaner with a colon
```scala
def pop_i32(): TypeVar.I32 => stack.pop().unbox_i32();
```

It would also be nice to have this for funexprs.

### DIFF
--- a/aeneas/src/vst/Parser.v3
+++ b/aeneas/src/vst/Parser.v3
@@ -513,7 +513,13 @@ component Parser {
 			var str = "this", pt = Parser.optKeyword(p, str);
 			if (pt != null) rettype = ReturnType.This(Token.new(p.fileName, str, loc.0, loc.1));
 			else rettype = ReturnType.Explicit(Parser.parseTypeRef(p));
-			body = if(p.curByte == ';', EmptyStmt.new(p.token(1)), Parser.parseBlockStmt(p));
+			if (Debug.UNSTABLE && p.enableSimpleBodies && (start = p.optN("=>")) >= 0) {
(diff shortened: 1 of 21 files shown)


mkhan45:
I think this also already adds explicit return types for funexprs so I should also add some tests there

mkhan45:
TODO to get this merged
- [ ] Fix `simple_body13.v3` - i just forgot some arrows
- [ ] Add a parser test with type params
- [ ] Add seman tests
- [ ] Add tests funexpr explicit rtp without `{...}`, in parser and seman

I wonder how much AI stuff would help

mkhan45:
I think it might've been faster to do myself but i probably had to think less

PR #464, opened 2025-08-04
[float] Minor cleanups and documentation improvements to Floats

titzer:
@JohnnyDingYQ I made a pass with some small cleanups and improvements to documentation.

### DIFF
--- a/lib/util/Floats.v3
+++ b/lib/util/Floats.v3
@@ -9,6 +9,7 @@ component Floats {
 	private def FLOAT_MANTISSA_BITS: byte = 23;
 	private def FLOAT_BIAS: byte = 127;
 
+	// Render {val} into the array {a} starting at {pos}.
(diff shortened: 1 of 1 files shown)


lib/util/Floats.v3
```diff
@@ -242,13 +226,12 @@ component Floats {
 
 		return length;
 	}
-
-	// if trailingZeros < 0, does not render decimal point
+	// If trailingZeros < 0, does not render decimal point.
```
JohnnyDingYQ: I forgot to clean up this comment from the time when `trailingZeros` was a `int` and a sentinel value. Currently, `trailingZeros` is a `u32`. This comment can be something like `// Render floating decimal in fixed notation with {trailingZeros} trailing zeros`

JohnnyDingYQ:
Looks good to me besides the one comment I forgot to remove.

PR #466, opened 2025-08-10
Fix minor bugs and improve developer experience

DavidLiedle:
This commit addresses several small issues found during code review:

Bug fixes:
- Fix indentation bug in NativeFileStream.putd() that caused early return for zero values
- The method now correctly prints '0' for zero input

Feature completions:
- Implement JSON support in PDK Host library (outputJsonValue, allocateJsonValue)
- Add missing lexer implementations in Parser library:
  - DfaLexer: Basic DFA state machine for pattern matching
  - DecimalLexer: Handles positive/negative decimal integers
  - StringLexer: Lexes quoted strings with escape sequence support

Developer experience improvements:
- Enhance parser error messages for better clarity:
  - "member expected" → "member name expected after '.' operator"
  - "identifier expected" → "identifier expected (must start with letter or underscore)"
  - "default match case expected" → "default match case expected (use '_' for default case)"
  - "',', or '=>' expected" → "',' to continue or '=>' to specify case body expected"

All changes have been tested with successful bootstrap and passing test suite.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/aeneas/src/vst/Parser.v3
+++ b/aeneas/src/vst/Parser.v3
@@ -289,7 +289,7 @@ component Parser {
 		var list: List<VstRepHint>;
 		while (p.curByte == '#') {
 			p.eat1();
-			if (!Char.isIdentStart(p.curByte)) p.error("identifier expected");
+			if (!Char.isIdentStart(p.curByte)) p.error("identifier expected (must start with letter or underscore)");
 			var d = p.input, q = p.curPos + 1;
 			while (q < d.length) {
 				var c = d[q];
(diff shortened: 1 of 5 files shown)


apps/Parser/TextParser.v3
```diff
@@ -43,7 +43,41 @@ class LexerChoiceCombo(lexers: Array<Lexer>) extends Lexer {
 
 // A graph-based deterministic finite automata lexer.
 class DfaLexer extends Lexer {
-	// TODO
+	// DFA implementation would require a state transition table
```
titzer, line 46: Given that `TextParser` is really just a demo application, nothing depends on it, so I don't think this change is a priority. I think I'd prefer this PR without this change.

aeneas/src/vst/Parser.v3
```diff
@@ -289,7 +289,7 @@ component Parser {
 		var list: List<VstRepHint>;
 		while (p.curByte == '#') {
 			p.eat1();
-			if (!Char.isIdentStart(p.curByte)) p.error("identifier expected");
+			if (!Char.isIdentStart(p.curByte)) p.error("identifier expected (must start with letter or underscore)");
```
titzer, line 292: Actually, it's not legal to start an identifier with an underscore.

PR #482, opened 2025-09-17
Use `#!/usr/bin/env bash` instead of `#!/bin/bash`

samestep:
[It's better](https://stackoverflow.com/a/10383546/5044950); for instance, `#!/bin/bash` doesn't work on NixOS.


titzer:
This needs conflict resolution; other than that I think it's good to go, nom :)

PR #489, opened 2025-10-08
Implement wfts (wrapped function type subsumption)

eliotmoss:
Suppose `B` is a subclass of `A` and there is a variable `f: void -> A`.
Further suppose `g` is a function of type `void -> B`.  By Virgil's rules
it is fine to assign `g` to `f`, but some targets, namely virtual machines
with their own subclass / type-checking rules, disallow this.  We can
work around the restriction by coding a _wrapper_ `w`:
```
def w() -> A {
  return g();
}
```
and assigning `w` to `f` instead.  A number of similar cases arise,
which can involve types of arguments to function as well as types
of results, nested use of function types, etc.  Further, assignments
between variables of function types (and implicit assignments in
passing arguments, returning results, etc.) imply the need to figure
out the wrapper to use at run time.  We do this my generating  a
_mapper_, which compares it argument (a function to wrap) against
all possible such functions and returning the wrapper for the matching
one.  Lastly, we also need _un_mappers to go from wrappers to their
underlying functions, in order to implement function comparison
correctly.  These mechanisms are _required_ for the Wasm GC target.
They also are required for the JVM target to support function
comparison.

From the standpoint of compiler internals, a key difference is making
sure that `TypeSubsume` (and `TypeCast`) operators that would normally
be considered no-ops are retained and suitable code is generated.

### DIFF
--- a/aeneas/src/ir/Normalization.v3
+++ b/aeneas/src/ir/Normalization.v3
@@ -14,6 +14,9 @@ class NormalizerConfig {
 	var NonRefClosureReceiver: bool;
 	var AnyRefOverflow: bool;
 	var NormalizeRange: bool;
+	var WrapFuncTypeSubsume: bool;
+	var ExplicitRefTypeCast: bool;
+	var FunctionCallReceiverTypeCast: bool;
 
 	var GetScalar: (Compiler, Program, Type) -> Scalar.set = defaultGetScalar;
 	var GetBitWidth: (Compiler, Program, Type) -> byte = defaultGetBitWidth;
@@ -39,12 +42,16 @@ def defaultGetBitWidth(compiler: Compiler, prog: Program, t: Type) -> byte {
 	}
 }
 
+def printWrap()      -> TerminalBuffer { return Reachability.printWrap(); }
+def printWrapExtra() -> TerminalBuffer { return Reachability.printWrapExtra(); }
+
 // Normalizes a program based on the results of reachability analysis.
-def TRANSFERRABLE_FACTS = (Fact.M_ABSTRACT | Fact.M_INLINE | Fact.M_ENUM_INIT | Fact.M_NEW | Fact.M_EMPTY | Fact.M_EQUALS);
+def TRANSFERRABLE_FACTS = (Fact.M_ABSTRACT | Fact.M_INLINE | Fact.M_ENUM_INIT | Fact.M_NEW | Fact.M_EMPTY | Fact.M_EQUALS | Fact.M_WRAPPER | Fact.M_MAPPER | Fact.M_UNMAPPER);
 class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer) {
 	def liveClasses = Vector<RaClass>.new();
 	def context = SsaContext.new(ra.compiler, ra.prog);
 	def typeMap = TypeUtil.newTypeMap<TypeNorm>();
+	def normQueue = WorkQueue.new();
 	var recordMap = V3.newRecordMap<Record>(); // XXX: canonicalize equivalent variant records
 	var complexRecordMap = V3.newRecordMap<Array<Val>>();
 	var newIr = IrModule.new();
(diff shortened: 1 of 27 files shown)


aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -264,6 +270,10 @@ class SsaRaNormalizer extends SsaRebuilder {
 				var t = extractMethodRef(orig, method), funcNorm = t.0, m = t.1;
 				var newOp = V3Op.newCallClassMethod(m);
 				var ai_new = normArgs(funcNorm, genRefs(i_old.inputs));
+				if (method.facts.M_NORM && V3.isComponent(method.receiver)) {
```
titzer: It shouldn't be possible to have a component method for `CallClassMethod`.

aeneas/src/ir/Normalization.v3
```diff
@@ -439,26 +886,35 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
...
-			normValIntoArray(old[i], etn, temp, 0);
+			normValIntoArray(old[i], enorm, temp, 0);
 			for (j < newRecords.length) {
 				newRecords[j].values[i] = temp[j];
 			}
 		}
 	}
 	// map a record 1-1
-	def normSimpleVal(tn: TypeNorm, v: Val) -> Val {
+	def normSimpleVal(tn: TypeNorm, t: Type, v: Val) -> Val {
```
titzer: I think if we can get by without adding a new parameter to this method by making sure the `tn` is not null in the case of field normalization.

aeneas/src/ir/Normalization.v3
```diff
@@ -427,7 +874,7 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 			var rf = rfs[i];
 			if (rf != null && rf.normIndices.0 >= 0) {
 				var v = oldVals[i];
-				if (rf.fieldType == null) newVals[rf.normIndices.0] = normSimpleVal(null, v);
+				if (rf.fieldType == null) newVals[rf.normIndices.0] = normSimpleVal(null, rf.getClosedType(), v);
```
titzer: You should be able to pass the typenorm of the field's type here. The `rf.fieldType == null` check is handling a prior optimization where simple closed field types that don't have a normalization are omitted.

aeneas/src/ir/Normalization.v3
```diff
@@ -439,26 +886,35 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 	}
 	// normalize the live instances of a complex (i.e. size-N element) array type
 	def normComplexArrayRecord(rt: ArrayNorm, oldRecord: Record, newRecords: Array<Record>) {
-		var etn = norm(V3Array.elementType(rt.oldType));
+		var enorm = norm(V3Array.elementType(rt.oldType));
```
titzer: There is already a field on `ArrayNorm` for the element normalization, which should be non-null for mixed arrays.

aeneas/src/ir/Normalization.v3
```diff
@@ -589,16 +1065,79 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
...
+			x: FuncVal => {
+				var fn = normalizeFuncRef(tn, x.memberRef);  // maybe wrap function
+				if (index < array.length) array[index] = fn;
+			}
 			_ => if (index < array.length) array[index] = v;
 		}
 	}
+	// Note: here, spec is the un-normalized IrSpec of the function
+	def normalizeFuncRef(tn: TypeNorm, spec: IrSpec) -> Val {
+		if (config.WrapFuncTypeSubsume) {
```
titzer: I think it'd be good to put the cast to `FuncNorm` in the caller. Also, returning early `if (!config.WrapFuncTypeSubsume)` will save a level of indentation and make it clear nothing else will happen when this is turned on.

aeneas/src/ir/Normalization.v3
```diff
@@ -589,16 +1065,79 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
...
 		// XXX: canonicalize normalized IrSpecs
 		var rm = spec.asMethod().raMethod;
 		var ta = spec.typeArgs;
 		if (rm == null) {
-			var rc = ra.getClass(spec.receiver);
-			rm = rc.findMethod(spec.member.index, ta);
+			var receiver = spec.receiver;
+			var rc = ra.getClass(receiver);
+			var meth = spec.asMethod();
+			if (rc == null) rc = ra.makeClass(receiver);
```
titzer: Did you find a case where the `RaClass` was null for a spec? Making a new `RaClass` at normalization time might indicate a problem, because there'll be an ordering problem with numbering classes and building their representations.

aeneas/src/ir/Reachability.v3
```diff
@@ -20,6 +20,18 @@ enum RaFact {
...
+component Reachability {
+	def printWrap() -> TerminalBuffer {
+		return Terminal.buf.purple().puts("Wrapping: ").end();
+	}
+	def printWrapExtra() -> TerminalBuffer {
+		return Terminal.buf.red().puts("Wrapping: ").end();
+	}
+}
+
+def printWrap()      -> TerminalBuffer { return Reachability.printWrap(); }
```
titzer: Can just be `def printWrap = Reachability.printWrap`

aeneas/src/ssa/SsaBuilder.v3
```diff
@@ -450,17 +451,90 @@ class SsaBuilder {
 	}
 	// TypeCast<F, T>(x)
 	def opTypeCast(ft: Type, tt: Type, x: SsaInstr) -> SsaInstr {
+		return opTypeCast0(ft, tt, x, null);
+	}
+	// opTypeCast0 allows indication of whether we are normalizing;
+	// when normalizing, we may request function wrappers / mappers
+	def opTypeCast0(ft: Type, tt: Type, x: SsaInstr, analyzer: ReachabilityAnalyzer) -> SsaInstr {
```
titzer, line 458: AFAICT this is only ever called with a non-null analyzer from with `SsaNormalizer`, so I think the wrapping-related functionality could be extracted out to that caller. Granted, there is a recursive invocation for the tuple case, but I think it would decouple the builder from the reachability / normalization.

aeneas/src/ssa/SsaBuilder.v3
```diff
@@ -542,25 +686,45 @@ class SsaBuilder {
...
+			FUNCREF, CLOSURE => {
+				if (context.compiler.NormConfig.WrapFuncTypeSubsume && ft != Null.TYPE) return opTypeCast0(ft, tt, x, analyzer);
 			}
 			_ => ;
 		}
 		if (SsaConst.?(x)) return graph.valConst(tt, SsaConst.!(x).val); // fold TypeSubsume(K)
  		return add(V3Op.newTypeSubsume(ft, tt), [x], x.facts);
 	}
+	// Version used when function wrapping/mapping may occur
+	def opNormFuncTypeSubsume(ft: Type, tt: Type, ftn: Type, ttn: Type, x: SsaInstr, analyzer: ReachabilityAnalyzer) -> SsaInstr {
```
titzer: Here, too, this is only called in `SsaNormalizer` and should probably live there.

aeneas/src/ir/Reachability.v3
```diff
@@ -351,6 +378,419 @@ class ReachabilityAnalyzer(compilation: Compilation) {
...
+			}
+			if (meth.typeArgs == null) {
+				Terminal.buf.puts("   meth typeArgs is null").outln();
+			} else {
+				Terminal.buf.put1("   meth typeArgs %q", meth.typeArgs.render).outln();
+			}
+		}
+
+		meth.setFact(Fact.M_WRAPPER);
+		var inner = SsaContext.new(Aeneas.compiler, prog);
```
titzer: `SsaContext` has a compiler.

aeneas/src/main/Compiler.v3
```diff
@@ -117,10 +117,16 @@ class Compiler(target: Target) {
...
 		NormConfig.NormalizeRange		= CLOptions.NR.get();
+		NormConfig.WrapFuncTypeSubsume		= CLOptions.WFTS.get();
+		NormConfig.AnyRefOverflow		= !CLOptions.WFTS.get();  // turn off when WFTS is on
+		NormConfig.ExplicitRefTypeCast		= CLOptions.WFTS.get();   // turn on when WFTS is on
+		NormConfig.FunctionCallReceiverTypeCast = false;
+
 		// target configures the compiler only once
 		if (target != null) target.configureCompiler(this);
+
+		Aeneas.compiler = this;
```
titzer: I think we can remove this global compiler reference.

aeneas/src/ssa/SsaOptimizer.v3
```diff
@@ -705,7 +713,11 @@ class SsaInstrReducer(context: SsaContext) extends SsaInstrMatcher {
 				}
 
 				match (cast) {
-					TRUE, SUBSUME => return x; // fold TypeCast
+					TRUE, SUBSUME => {
```
titzer, line 716: Looks like an incomplete migration in the existing code to the new method `TypeSystem.computeCast` that will distinguish between functions and other subsumptions. I think this is OK for now but could bear a TODO.

aeneas/src/ssa/SsaOptimizer.v3
```diff
@@ -1070,6 +1088,38 @@ class SsaInstrReducer(context: SsaContext) extends SsaInstrMatcher {
 					if (d == null) return addThrow(i.source, V3Exception.NullCheck);
 					// CallFunction(#K) => CallMethod
 					var inputs = copyInputs(i, null, 1);
+
```
titzer: This looks pretty tricky and should probably have the functionality to adapt arguments extracted to a helper function.

aeneas/test/MachStackifierTest.v3
```diff
@@ -13,14 +13,14 @@ def ARCH_TEE = 5;
 def ARCH_POP = 6;
 def ARCH_CONST = 7;
 
-def prog = Program.new();
+def prog = Program.new(null);
```
titzer: I think you added and then removed a parameter here.

aeneas/src/ir/SsaNormalizer.v3
eliotmoss: Verified that that code is unnecessary.  Will delete.

aeneas/test/MachStackifierTest.v3
eliotmoss: Yes.  I have fixed that back.

aeneas/src/ir/Normalization.v3
eliotmoss: Ok, I moved the cast to the several callers and flipped the sense of the "if".  (Note: that required duplicating the last two lines of the method.)

aeneas/src/ir/Normalization.v3
eliotmoss: Yes, but apparently null (at least sometimes) for _complex_ arrays (which I gather are arrays of boxed records while mixed arrays are arrays of unboxed records).  This is apparent from `visitArray`.  Note that the only change I made was the name of the variable, which did seem arbitrary so I changed it back.

aeneas/src/ir/Reachability.v3
eliotmoss: Yes; this is how I was setting it when creating this "inner" context.  Maybe you meant to say that ReachabilityAnalyzer has a compiler field, which I should have used here :-) .  I have made that change.  At the time I was not seeing another way.

aeneas/src/ir/Normalization.v3
eliotmoss: Yes, it seems there are a number of possibly-null cases, but `normSimpleVal` doesn't care about them :-) ...

aeneas/src/ir/Normalization.v3
eliotmoss: Yes, I rewrote to use `norm(rf.getClosedType())`.

aeneas/src/ir/Reachability.v3
eliotmoss: Yes, cleaner.

aeneas/src/main/Compiler.v3
eliotmoss: Done.

aeneas/src/ir/Normalization.v3
eliotmoss: Must have been some interim problem.  Things work fine now without that line, so I deleted it.

aeneas/src/ssa/SsaBuilder.v3
eliotmoss: Good catch!  I moved this as suggested.

aeneas/src/ssa/SsaBuilder.v3
eliotmoss: I moved this one also.

aeneas/src/ssa/SsaOptimizer.v3
eliotmoss: Not sure that migration would help much, but I added the TODO :-)

aeneas/src/ssa/SsaOptimizer.v3
eliotmoss: Moved the code to a helper function as suggested, and found some code that was not doing anything and could be deleted.

aeneas/src/ir/Normalization.v3
```diff
@@ -1066,53 +1064,54 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
...
+				spec = ra.genWrappedFunction(ft, tt, normed.1, norm(normed.0.receiver));
+				if (CLOptions.PRINT_WRAP.get()) {
+					printWrap().put2("normalizeFuncRef: ft = %q  tt = %q", ft.render, tt.render)
+							.put3("  meth = %q{%d/%d}", meth.render, meth.index, meth.uid)
+							.outln();
+				}
+				// insure the wrapper is visited, etc.
+				var rm = ra.makeMethod(spec.typeArgs, IrMethod.!(spec.member), null);
+				var rc = ra.getClass(rm.receiver);
+				ra.getMethod(null, rm);
```
titzer: I was just looking at this again, because it wasn't obvious to me. `getMethod()` has the side effect of potentially adding the method to the reachability queue, unless the method is already marked as `RM_LIVE`. Which of the side effects are needed here?

aeneas/src/ssa/SsaOptimizer.v3
```diff
@@ -1069,8 +1088,8 @@ class SsaInstrReducer(context: SsaContext) extends SsaInstrMatcher {
 					var d = FuncVal.!(xval);
 					if (d == null) return addThrow(i.source, V3Exception.NullCheck);
 					// CallFunction(#K) => CallMethod
-					var inputs = copyInputs(i, null, 1);
-					return add(i.source, V3Op.newCallMethod(d.memberRef), inputs).setFact(Fact.O_NO_NULL_CHECK);
+					adaptKnownFunctionInputs(i, d);  // add any needed subsumptions / casts
```
titzer, line 1091: I'm curious if we need this adaptation for devirtualization as well?

aeneas/src/ssa/SsaOptimizer.v3
```diff
@@ -1155,6 +1174,37 @@ class SsaInstrReducer(context: SsaContext) extends SsaInstrMatcher {
...
+				.put2("  call type %q.%q", AnyRef.TYPE.render, ftype.render).outln();
+		}
+		var changed = false;
+		if (xtype != ftype) {
+			var inputs = Ssa.inputs(i);
+			for (j = 1; j < xsig.paramTypes.length; ++j) {
+				var input = inputs[j];
+				var ft = fsig.paramTypes[j], tt = xsig.paramTypes[j], atype = input.getType();
+				if (SsaConst.?(input)) atype = SsaConst.!(input).vtype;
+				if (ft != tt && atype != tt) {
```
titzer: IIUC I think we can always use a TypeSubsume, since these should never fail.

aeneas/src/ssa/SsaOptimizer.v3
```diff
@@ -1155,6 +1174,37 @@ class SsaInstrReducer(context: SsaContext) extends SsaInstrMatcher {
...
+		var xsig = FuncType.!(xtype).sig();
+		var rcvr = d.memberRef.asMethod().receiver;
+		if (CLOptions.PRINT_WRAP_EXTRA.get()) {
+			var in0type = i.inputs[0].dest.getType();
+			printWrapExtra()
+				.put3("SsaOptimizer: adaptCallKnownFunctionInputs const type %q.%q  arg type %q", rcvr.render, xtype.render, in0type.render)
+				.put2("  call type %q.%q", AnyRef.TYPE.render, ftype.render).outln();
+		}
+		var changed = false;
+		if (xtype != ftype) {
```
titzer: I think I like an early return better; it saves indentation and makes it clear nothing happens afterward.

aeneas/src/ssa/SsaOptimizer.v3
```diff
@@ -1155,6 +1174,37 @@ class SsaInstrReducer(context: SsaContext) extends SsaInstrMatcher {
...
+				if (ft != tt && atype != tt) {
+					// add subsume/cast if argument type is different from the static type
+					changed = true;
+					if (TypeSystem.isSubtype(atype, tt))
+						inputs[j] = add(i.source, V3Op.newTypeSubsume(atype, tt), [input]);
+					else
+						inputs[j] = add(i.source, V3Op.newTypeCast(atype, tt), [input]);
+				}
+			}
+			if (changed) i.setInputs(inputs);
```
titzer: Any reason not to just update each dataflow edge in place as opposed to replacing them all at once?

aeneas/test/MachStackifierTest.v3
```diff
@@ -81,7 +81,7 @@ class StackOpInstrBuffer extends ArchInstrBuffer {
 	new(codegen: SsaMachGen, prog: Program, regSet: MachRegSet) super(codegen, prog, regSet) { }
 	def putArchInstr(indent: int, i: ArchInstr) -> int {
 		var opcode = int.view(i.opcode()), a = i.operands;
-		var name = WasmOpNames.array[opcode];
+//		var name = WasmOpNames.array[opcode];
```
titzer: I guess we didn't need this line after all :-)

aeneas/src/ir/Normalization.v3
eliotmoss: Turns out only the marking for liveness and adding to the `liveMethods` vector are needed.  I recoded to:

```
    if (!rm.setFact(RaFact.RM_LIVE)) {
            ra.liveMethods.put(rm);
    }

```

aeneas/src/ssa/SsaOptimizer.v3
eliotmoss: Yes, because the specific method implementation may have a signature that is a subtype of the virtual method signature, and it may be because of function subtypes.  (Assuming we're talking about the same thing :-) .)

aeneas/src/ssa/SsaOptimizer.v3
eliotmoss: Yes, I verified that you're right, and updated the code for that.

aeneas/src/ssa/SsaOptimizer.v3
eliotmoss: Yes, that's better.

aeneas/src/ssa/SsaOptimizer.v3
eliotmoss: I changed it to make the change in place using `SsaDfEdge.new(...)`.

aeneas/test/MachStackifierTest.v3
eliotmoss: I think it was added for debugging at some point.  I'm removing it now.

aeneas/src/ssa/SsaOptimizer.v3
titzer: There is `Edge.update()`; otherwise there'll be an orphaned `SsaDfEdge` hanging out in the useList.

eliotmoss: Ah - Edge is a library I was less familiar with ... fixed it.

PR #510, opened 2025-11-12
[x86-64] Fix improper instruction selection for subtraction

matthew-mojira:
The following program is incorrect on `x86-64-linux` (with `-O0`):
```
def main() -> int {
    def x: long = 0;
    return int.view(1 - x);
}
```
It returns 0, instead of 1, because the assembly generator only checks the constant `1` against an int.

### DIFF
--- a/aeneas/src/x86-64/SsaX86_64Gen.v3
+++ b/aeneas/src/x86-64/SsaX86_64Gen.v3
@@ -123,7 +123,10 @@ class SsaX86_64Gen extends SsaMachGen {
 			}
 			IntSub => {
 				var yval = m.intbinop(i);
-				if (MATCH_NEG && m.xconst && m.xint == 0) return emit1(I_NEGQ | (AM_OP << AM_SHIFT), ovwReg(i, m.y));
+				if (MATCH_NEG && m.xconst) {
+					if (m.inttype.width <= 32 && m.xint == 0) return emit1(I_NEGQ | (AM_OP << AM_SHIFT), ovwReg(i, m.y));
+					if (m.inttype.width > 32 && m.xlong == 0) return emit1(I_NEGQ | (AM_OP << AM_SHIFT), ovwReg(i, m.y));
+				}
 				emitIntBinop(I_SUBD, i);
 			}
 			IntMul => {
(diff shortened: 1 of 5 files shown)


aeneas/src/x86-64/SsaX86_64Gen.v3
```diff
@@ -123,7 +123,10 @@ class SsaX86_64Gen extends SsaMachGen {
 			}
 			IntSub => {
 				var yval = m.intbinop(i);
-				if (MATCH_NEG && m.xconst && m.xint == 0) return emit1(I_NEGQ | (AM_OP << AM_SHIFT), ovwReg(i, m.y));
+				if (MATCH_NEG && m.xconst &&
```
titzer, line 126: Is it possible to emit a `NEGQ` or `NEGD` depending on the operand size?

test/core/int_sub05.v3
```diff
@@ -0,0 +1,6 @@
+//@execute 0=1; 1=0
+component int_sub05 {
```
titzer, line 2: You can omit the `component` and just have a top-level main.

test/core/long_sub01.v3
```diff
@@ -0,0 +1,4 @@
+//@execute (0,1)=-1; (1,-1)=2; (-200,13)=-213
+component long_sub01 {
+	def main(a: int, b: int) -> int { return int.view(long.!(a) - long.!(b)); }
```
titzer, line 3: Can you also write a version of these tests that just have the parameter type of `long`?

matthew-mojira: I was unable to run the tests for that on `js@node` because of:
```
long_sub01: fail: Run 0 failed: TypeError: Cannot convert 0 to a BigInt
```
Also, I can't run it with the return value as `long` because of (on `v3i`):
```
long_sub01.v3: fail: (0, 1)=-1L, expected: -1
```
and then I run into a parse error when updating the test expectations to include the `L`.

titzer: Ok, yeah, I forgot about that. Can you factor it internally to have a function that has a long parameter? (That is so that the `long.view` operation is not involved).

aeneas/src/x86-64/SsaX86_64Gen.v3
matthew-mojira: #690

PR #511, opened 2025-11-12
`Ranges.v3` quicksort bug fix

linxuanm:
This PR fixes the quicksort in `Ranges.v3`.

### DIFF
--- a/test/lib/RangesTest.v3
+++ b/test/lib/RangesTest.v3
@@ -6,6 +6,7 @@ def X = [
 	T("reverse_int", test_reverse_int),
 	T("reverse_string", test_reverse_string),
 	T("sort_ints", test_sort_ints),
+	T("sort_ints_4", test_sort_ints_4),
 	T("sort_floats", test_sort_floats),
 	T("sort_strings", test_sort_strings),
 	T("sort_perm3", test_sort_perm3),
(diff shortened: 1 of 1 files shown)


lib/util/Ranges.v3
```diff
@@ -51,31 +51,29 @@ component Ranges {
...
-				i++;
-			} else {
-				r[i] = r[j];
-				r[j] = a;
-				j--;
-			}
-		}
-		return i;
-	}
+    if (r.length <= 1) return;
```
titzer: Looks like spurious formatting changes.

lib/util/Ranges.v3
```diff
@@ -51,31 +51,29 @@ component Ranges {
...
+    }
+    def pi = partition(r, cmp);
+    if (pi > 1) quicksort(r[0 ... pi], cmp);
+    if (pi < r.length - 2) quicksort(r[pi + 1 ...], cmp);
+  }
+  private def partition<T>(r: Range<T>, cmp: (T, T) -> bool) -> int {
+    def pivot = r[r.length - 1];
+    var i = 0;
+    for (j = 0; j < r.length; j++) {
+      if (cmp(r[j], pivot) || j == r.length - 1) {
```
titzer: Is it possible that cmp(r[j], pivot) is never true, so only the last iteration will increment i to 1, and then the return at the end will return i-1 == 0?

Can you write a test that tests all possible 4-length arrays containing the elements 1, 2, 3, 4? E.g. [1, 1, 1, 1], [1, 1, 1, 2] etc? I think that will be guaranteed to hit all possible cases of partition.

lib/util/Ranges.v3
linxuanm: Oops sorry I was trying out a new VS Code profile, must've messed up the tab/space indentation.

lib/util/Ranges.v3
linxuanm: I think the case you mentioned still works, since if `cmp(r[j], pivot)` is never true, then `pivot` is less than all other `r[j]` and thus should be at index `0` when the function returns.

titzer (approved):
lgtm, we cross our fingers...

PR #520, opened 2025-11-20
Packing bool and enumset cases

dnguyen-23:
Added cases to SsaPacker to handle packing bools and enumsets. Tests for packing bools, enumsets, tuples of bools, and tuples of enumsets are included in this PR.


titzer (approved):
lgtm

PR #542, opened 2025-12-19
Adding support for exports and import…

eliotmoss:
Add support for producing linkable .o files on x86-64-linux, for imports and exports.  Also adds tests and tutorial documentation.


aeneas/src/mach/MachProgram.v3
```diff
@@ -21,6 +21,7 @@ class MachProgram extends TargetProgram {
 	def fieldMap      = IrUtil.newIrItemMap<Address<IrField>>();
 	def recordMap     = V3.newRecordMap<Address<Record>>();
 	def recordQueue   = Vector<Record>.new();
+	var importLocs: HashMap<long, List<(long, int)>>;
```
titzer: I'd like to avoid another data structure, especially another hash map that needs to be looked at frequently. Can we encode this as negative absolute addresses?

aeneas/src/os/Example.v3.save
```diff
@@ -0,0 +1,159 @@
+class C {
```
titzer: Not sure if this is a spurious check-in, it's named `.save`.

aeneas/src/os/Linux.v3
```diff
@@ -149,6 +160,54 @@ class LinuxTarget extends Target {
...
+		var importNames = HashMap<string, void>.new(Strings.hash, Strings.equal);
+		// Map from addresses of imported functions to their symbol table indexes
+		var importAddrs = HashMap<long, int>.new(int.view, long.==);
+		// Map from addresses of imported functions to locations that refer to them;
+		// locations are (virtual address, offset in DataWriter)
+		var importLocs = HashMap<long, List<(long, int)>>.new(int.view, long.==);
+		// Flag to track whether a symbol section is needed
+		var anySymbols = false;
+		if (haveImports) {
+			for (i < prog.ir.classes.length) {
```
titzer: I think we can probably avoid another iteration over all the components if we roll this traversal into `MachProgram`--it'll also be more target independent that way.

aeneas/src/os/Linux.v3
```diff
@@ -190,9 +249,114 @@ class LinuxTarget extends Target {
...
+			if (emitRel) {
+				var tempAddr = Addr.new(null, null, 0);
+				tempAddr.absolute = code.p_vaddr;
+				sections.addAbsSymbol(w, ".text.virgil.base", tempAddr);
+				tempAddr.absolute = data.p_vaddr;
+				sections.addAbsSymbol(w, ".data.virgil.base", tempAddr);
+			}
 
-			if (CLOptions.SYMBOLS.val) {
+			def exportAddrs = HashMap<string, Addr>.new(Strings.hash, Strings.equal);
```
titzer: Seems like this should also probably live in the `MachProgram` and only be allocated if named exports are found.

aeneas/src/os/Linux.v3
```diff
@@ -190,9 +249,114 @@ class LinuxTarget extends Target {
...
+				var mainAddr = mach.methodMap[mainMeth];
+				if (mainAddr != null && mainAddr.absolute >= 0) {
+					exportAddrs["main"] = mainAddr;
+					anySymbols = true;
+					sections.addFuncSymbol(w, "main", mainAddr);
+				}
+			}
+
+			if (haveImports) {
+				for (node = w.patches; node != null; node = node.tail) {
```
titzer: I think this might make more sense as a callback from the `MachDataWriter` when it patches an address which is imported/relocatable, rather than iterating over the patches separately.

aeneas/src/os/Linux.v3
```diff
@@ -240,21 +411,31 @@ enum DebugSection(printName: string) {
...
+class ElfSections(elf: ElfHeader, haveExports: bool, haveImports: bool, debugSymbol: bool) {
 	def str = ElfSectionHeader.new();
 	def sym = ElfSectionHeader.new();
 	def text = ElfSectionHeader.new();
 	def data = ElfSectionHeader.new();
 	def debugs = Array<ElfSectionHeader>.new(DebugSectionCount);
 	def symbols = Vector<ElfSymbol>.new();
-
+	def relocatable = haveExports | haveImports;
+	def text_relocs = ElfSectionHeader.new();
```
titzer: Having relocations is going to be an uncommon case. Can we make these lazy instead?

aeneas/src/os/Linux.v3
```diff
@@ -330,28 +541,140 @@ class ElfSections(elf: ElfHeader, debugSymbol: bool) {
 		}
 		sym.sh_size = w.pos - sym.sh_offset;
 	}
-	def addFuncSymbol(w: MachDataWriter, str: string, addr: Addr) {
+	def encodeReloc32(w: DataWriter, reloc: RelocInfo) {
+		match (reloc.rtype) {
+			Elf32_R_Type.R_386_32.code => {
```
titzer: The specific encoding of ELF relocations should be with the other ELF code in `ELF.v3`.

aeneas/src/vst/Parser.v3
```diff
@@ -131,7 +131,7 @@ component Parser {
 			'i' => {
 				var importToken = optKeyword(p, "import");
 				if (importToken != null) {
-					var importName = parseOptString(p);
+					var importName = stripQuotes(parseOptString(p));
```
titzer: I think it's better to keep the original token and then evaluate it later in the compiler. In particular, it needs more than just stripping the quotes, because of escapes.

aeneas/src/vst/Verifier.v3
```diff
@@ -231,24 +233,29 @@ class Verifier(compiler: Compiler, prog: Program) {
...
+			var prev = exportMap[decl.expname.image];
 			if (prev != null) {
-				cv.errAtDecl(decl).ExportDeclError(formatPrev("export redefined", prev.token));
+				cv.errAtToken(decl.expname).ExportDeclError(formatPrev("export redefined", prev.token));
 				return;
 			}
 		}
-		exportMap[decl.token.image] = decl;
+
+		exportMap[decl.expname.image] = decl;
```
titzer: This should probably be the evaluated literal, not just the token's image.

aeneas/src/vst/Verifier.v3
```diff
@@ -631,6 +638,14 @@ class VstCompoundVerifier {
...
 				if (decl.importName == null) decl.importName = decl.token;
 				if (decl.typeParams != null) errAtDecl(decl).ImportError("imported methods cannot have type parameters");
+
+				var compName = VstComponent.!(decl.receiver).importName.image;
+				var sep = if(compName.length == 0, "", ".");
+				var fullName = Strings.builderOf(compName).puts(sep).puts(decl.importName.image).extract();
+
+				if (verifier.importNames == null) verifier.importNames = HashMap<string, void>.new(Strings.hash, Strings.equal);
+				else if (verifier.importNames.has(fullName)) errAtToken(decl.importName).ImportError("duplicate import name");
+				verifier.importNames[fullName] = ();
```
titzer: Should the import name be part of the IrRoot? that might obviate the need for yet another map.

aeneas/src/mach/MachProgram.v3
eliotmoss: Accomplished.

aeneas/src/os/Example.v3.save
eliotmoss: You're right - I'll remove it.

aeneas/src/vst/Verifier.v3
eliotmoss: That might be better.  I was avoiding changing things in the Ir.

aeneas/src/vst/Parser.v3
titzer: Ah, looks like `parseOptString` discards the evaluated literal (the bytes). `parseStringExpr` doesn't. The magic is in `lib/util/Strings.parseLiteral`.

eliotmoss: Ok - what I have now coded uses the result of `Strings.parseLiteral` if the token image starts with a `"`, and otherwise just the token image.

aeneas/src/vst/Verifier.v3
eliotmoss: So, for the "evaluated literal" cases what I did was add a new method to `Parser`, `canonExternalName: Token -> string`.  This uses `Strings.parseLiteral` to parse the image if it starts with `"` and otherwise just returns the image.

aeneas/src/vst/Verifier.v3
eliotmoss: We don't have an `IrRoot` in hand here.  The map (set, really) is constructed lazily and is used only to check for duplicate names.  But the main point is that since one can give an arbitrary explicit string name to an import, arbitrary clashes can occur.  So, if I have an imported method in hand, how do I check that no other method has been imported under the same name?  This set is also used to make sure that no export name clashes with an import name.

aeneas/src/os/Linux.v3
eliotmoss: Refactored.

aeneas/src/vst/Verifier.v3
eliotmoss: I combined `importNames` and `exportMap` into just `exportMap`, by having entries for imports map to `null`.  With more work we could get slightly better error messages in some cases (being able to indicate the import that conflicts with a later import/export), but I'm not sure it's worth it.

aeneas/src/os/Linux.v3
eliotmoss: Done.

aeneas/src/os/Linux.v3
eliotmoss: To make that general enough to be able to handle full relocation in the future, I am thinking of adding to `Addr` a flags `EnumSet` and having a `MachDataWriter` patch call that accepts a callback function and a flags mask to indicate which patches need a callback.  WDYT?

aeneas/src/os/Linux.v3
eliotmoss: I got it into `MachProgram`, but it needs to be a separate iteration so that imports' addresses can be assigned before any other method tries to refer to them.

aeneas/src/os/Linux.v3
eliotmoss: I did not make that proposed change, but did a less general callback.

aeneas/src/os/Linux.v3
eliotmoss: Moved.

PR #548, opened 2025-12-27
[packing] Unpack operator implemented with support for primitive types, tuples, and ADTs with no variants. 

dnguyen-23:
This PR features changes to X86CodeGen for gen_sse_view_int where uses should be processed through the sse function. Test cases have been added to test the functionalities implemented in this PR (see in virgil/test/variants).

### DIFF
--- a/aeneas/src/ir/Packing.v3
+++ b/aeneas/src/ir/Packing.v3
@@ -2,6 +2,16 @@
 // See LICENSE for details of Apache 2.0 license.
 
 // NOTE: For classes and tuples, first elements at higher bits, last elements at lower bits. 
+
+// Packs data into an IntType. 
+// Given a `ft` (fromType), `flatten` retreives each scalar 
+// from `input`, based on its type, `ft`. Scalars are stored in  
+// `scalars` and then packed into the final result through `finish`.
+// 
+// Note on the order scalars are packed: 
+// For types like tuples and ADTs with multiple scalars, 
+// the first elements are packed into the higher order bits 
+// The last elements are stored at the least significant bits 
 // Intuitively, convention allows us to read the class as a binary number 0b.....
 class SsaPacker(builder: SsaBuilder, result: IntType) {
 	def scalars = Vector<(PrimType, int, SsaInstr)>.new();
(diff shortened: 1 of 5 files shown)


aeneas/src/ir/Packing.v3
```diff
@@ -1,6 +1,8 @@
 // Copyright 2025 Virgil authors. All rights reserved.
 // See LICENSE for details of Apache 2.0 license.
 
+// NOTE: For classes and tuples, first elements at higher bits, last elements at lower bits. 
+// Intuitively, convention allows us to read the class as a binary number 0b.....
```
titzer, line 15: The comments for a class should describe what the class is for and its main responsibility.

aeneas/src/ir/Packing.v3
```diff
@@ -72,5 +75,129 @@ class SsaPacker(builder: SsaBuilder, result: IntType) {
...
+class SsaUnpacker(builder: SsaBuilder, ft: IntType) {
+	// Track the position in the input IntType we are unpacking
+	var pos: int = 0;
+	// A buffer to hold the scalars so that we can construct tuples and variants. 
+	def unpackedBuffer = ListStack<SsaInstr>.new();
+	var input: SsaInstr;
+	def unpack(tt: Type, i: SsaInstr) -> SsaInstr {
+		input = i;
+		unpack0(tt);
+		// if (unpackedBuffer.empty() ) { builder.context.fail(Strings.format1("Error unpacking %q. Nothing unpacked", tt.render)); }
```
titzer: Any commented-out code should be removed.

aeneas/src/ir/Packing.v3
```diff
@@ -72,5 +75,129 @@ class SsaPacker(builder: SsaBuilder, result: IntType) {
...
+				def fieldInstr = Array<SsaInstr>.new(fs.length);
+				for (i < fs.length) {
+					if (unpackedBuffer.empty()) { 
+						builder.context.fail(Strings.format2("Error unpacking %q. Missing %q", x.render, fs[i].render));
+					}
+					fieldInstr[i] = unpackedBuffer.pop();
+				}
+
+				def irFieldTypes = Arrays.map(fs, IrField.fieldType);
+				def variantAlloc = builder.add(V3Op.newVariantAlloc(x, irFieldTypes), fieldInstr, Facts.NONE);
```
titzer: You can use `builder.pure()`.

test/variants/bool_unpack03.v3
```diff
@@ -0,0 +1,38 @@
+//@execute false=true; true=true
+def main(b: bool) -> bool {
```
titzer: What I've generally done for tests like these is to use a return type of `int` and then return `-11`, `-12`, etc for each subsequent case. It makes easier to debug in case it should fail, because then you have a return value rather than working backward from a line number, which won't be available by default (without the runtime).

titzer (commented):
Generally looks good with minor comments. Awesome number of tests, I love it!

titzer (approved):
lgtm, thanks!

PR #631, opened 2026-04-01
Add open enums and enum methods with tests

eliotmoss:
This combines the open enums work into a single commit.  There were few minor conflicts, which have been worked out.  Passes all CI locally.

### DIFF
--- a/aeneas/src/core/Eval.v3
+++ b/aeneas/src/core/Eval.v3
@@ -854,8 +854,8 @@ def evalOp(op: Operator, args: Arguments) -> Result {
 //----------------------------------------------------------------------------
 		NullCheck => {
 			var val = args.vals[0];
-			if (!Record.?(val)) return Values.BOTTOM; // non-ref values (e.g. enums) are never null
 			if (val == null) return args.throw(V3Exception.NullCheck, null);
+			if (!Record.?(val)) return Values.BOTTOM; // non-ref values (e.g. enums) are never null
 			return Values.BOTTOM;
 		}
 		BoundsCheck => {
--- a/aeneas/src/core/Eval.v3
+++ b/aeneas/src/core/Eval.v3
@@ -126,6 +126,12 @@ component Eval {
 				return b == null;
 			}
 			CLASS_QUERY, VARIANT_QUERY => {
+				if (EnumType.?(tt)) {
+					// enum subtype query: check integer tag is in [enumTagLo, enumTagHi]
+					var decl = EnumType.!(tt).enumDecl;
+					var tagVal = Int.unbox(val);
+					return tagVal >= decl.enumTagLo && tagVal <= decl.enumTagHi;
+				}
 				if (val == null) return false;
 				if (tt.open()) return false;
 				return Record.?(val) && TypeSystem.isSubtype(Record.!(val).rtype, tt);
@@ -811,12 +817,22 @@ def evalOp(op: Operator, args: Arguments) -> Result {
 			return if(object != null, object.values[field.index]);
 		}
 		VariantGetMethod(method) => {
+			var ta = args.getTypeArgs();
+			if (EnumType.?(ta[0])) {
+				var spec = lookupEnumVirtual(args, method);
+				return Closure.new(args.vals[0], spec);
+			}
 			var object = args.r(0);
 			var spec = args.getClosedIrSpec(method);
 			if (spec == null) return args.notFoldable(null);
 			return Closure.new(object, spec);
 		}
 		VariantGetVirtual(method) => {
+			var ta = args.getTypeArgs();
+			if (EnumType.?(ta[0])) {
+				var spec = lookupEnumVirtual(args, method);
+				return Closure.new(args.vals[0], spec);
+			}
 			var object = getRecordReceiver(args);
 			var spec = args.getClosedIrSpec(method);
 			if (spec == null) return args.notFoldable(null);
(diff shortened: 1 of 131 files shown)


aeneas/src/core/Eval.v3
```diff
@@ -126,6 +126,12 @@ component Eval {
 				return b == null;
 			}
 			CLASS_QUERY, VARIANT_QUERY => {
+				if (EnumType.?(tt)) {
```
titzer, line 129: From this it's clear to me that enums need to be desugared to variants earlier in a more general way. Claude has just smeared special cases from front to middle to backend. There shouldn't be any need to alter the evaluation of operators like this.

aeneas/src/ir/Ir.v3
```diff
@@ -412,7 +412,9 @@ class IrModule {
 		return ic;
 	}
 	def newIrClass(ctype: Type, superClass: IrClass, decl: VstCompound) -> IrClass {
-		var ic = IrBuilder.new(ctype, superClass).buildClass(decl);
+		var builder = IrBuilder.new(ctype, superClass);
```
titzer: I don't know why it feels the need to add the IR module to the builder now.

aeneas/src/ir/Normalization.v3
```diff
@@ -158,6 +158,10 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 		if (rc.isUnboxed()) {
 			// move flattened data type receiver to function sig
 			ftype = Function.prependParamTypes(rc.variantNorm.sub, ftype);
+		} else if (EnumType.?(rc.oldType)) {
```
titzer, line 161: If enums are properly desugared to variants, this will just fall out of the normal unboxing.

aeneas/src/ir/Normalization.v3
```diff
@@ -588,11 +592,15 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 			rm.norm.flags |= IrFlag.M_OVERRIDE;
 			sm.norm.flags |= IrFlag.M_OVERRIDDEN;
 		}
+		// For enum methods, M_OVERRIDDEN must be transferred from original
```
titzer, line 595: I don't know why it thinks this.

aeneas/src/ir/Normalization.v3
```diff
@@ -630,6 +638,37 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 			setMtable(l.head, rv);
 		}
 	}
+	def layoutEnumMtable(rv: RaVirtual, rm: RaMethod, rc: RaClass) {
```
titzer, line 641: More replications with special cases from our AI friends. As this is very tricky, we definitely don't want to do this again.

aeneas/src/core/Eval.v3
eliotmoss: I'm willing to work on simplifying such things.  I'm not 100% clear on the desugaring you're hoping for, though.  I can see that enums are kind  of like variants that have no variant fields.  Of course enums have their own "fields", that are implemented as arrays indexed by tag.  If the two are to be unified into a single whole, we need some place to hang both kinds of fields.  Also while enum class structure mirrors variant class structure (in the case of no generic parameters), enums themselves are single values (or _have_ unique single values associated with them) while variant cases are types (unless they have no fields, in which case they have the same "ambiguity" as enums).  If you have a little more guidance I would be more confident revising code ...

titzer: Ah, forgive me. It used to be that enums were immediately desugared to a single class declaration. At some point I split off the implementation entirely. If open enums were desugared into open variants instead (i.e. adding a class per enum case), then enum method dispatch would just be variant dispatch. Enum fields would still be implemented via field arrays and use `VariantGetTag` as the index. Unboxing will just eliminate the overhead of boxing enum values.

eliotmoss: I get the idea but am foggy as to the particulars.  Specifically I wonder about the field arrays, and maybe also enumsets.  Is the desugaring proposed to happen in `VstSsaGen`, so that enums are mostly washed away once we have SSA, but the distinction is more visible in parsing (of course) and verification?  This would make sense in that, once verified, enums could be mapped down to other structures (variants, field arrays) and marked to be unboxed (always).  I'd have to get more deeply into it to see where issues might come up with enumsets, but again, suitable use of `VariantGetTag` would get the raw numeric values required.  So, I could try to combine processing in verification (and maybe parts of parsing), but strive to eliminate separate processing once in SSA.

aeneas/src/ir/Ir.v3
eliotmoss: Apparently the field and this code are dead, so I'm removing it.

aeneas/src/ir/Normalization.v3
eliotmoss: `isUnboxed` means that normalization did the unboxing - but enums are implicitly already unboxed, so the test doesn't work for them.  Unifying these more would be a bigger change.  Is that how you would prefer to go?

aeneas/src/ir/Normalization.v3
eliotmoss: Ultimately this appears necessary because enums don't use regular vtables so their overrides are processed differently.  That don't use regular vtables because individual enum cases are not types and thus don't have a corresponding `RaClass`, etc.  Again, is this deeper change one you would prefer to happen?

aeneas/src/ir/Normalization.v3
eliotmoss: Part of the same effects from enum cases not having an `RaClass` and thus not lining up with the existing dispatch mechanisms.

PR #644, opened 2026-04-21
Add LinuxConst.MAP_32BIT

matthew-mojira:


titzer:
Going to close this as applications should not rely on the Virgil runtime for Linux syscall definitions.

PR #650, opened 2026-04-24
Use VstOperator to implement VstNew instead of a VarBinding

titzer:


titzer:
I'll redo this after descriptors are fully working on all targets.

PR #651, opened 2026-04-28
Open enums A (no per-enum-case methods)

eliotmoss:
Open enums with subtypes and methods (but not methods on individual enum cases)


eliotmoss:
I'll see about adding a version of this as a single squashed commit.

The last two commits show the diff from Strategy B.

PR #652, opened 2026-04-28
Open enums B (with per-enum-case methods)

eliotmoss:
Open enums with subtypes and methods, including methods on individual enum cases


eliotmoss:
As with A, I'll see about preparing a version with the commits squashed together.

Issue #649, opened 2026-04-22
unboxed variant causes compiler failure


terry3800:
These type definitions: 

```
type Value {
	case Ref(val: ObjRef);
	case I31(val: u31);
	case I32(val: u32);
	case I64(val: u64);
	case F32(bits: u32);
	case F64(bits: u64);
	case V128(low: u64, high: u64);
	case Cont(val: Continuation);
	//case Wide(val: WideRef); // create three files to declare different kind of widerefs: object, u64, (object + u64)
}

type ObjRef #unboxed {
	case Heap(val: Object);
	case External(val: u63);
}
```

result in this compiler failure: 

```
./build.sh  wizeng x86-64-linux
./build.sh unittest x86-64-linux
!NullCheckException
        in VariantNorm.tagType() [aeneas/src/ir/VariantNormalizer.v3 @ 42:56]
        in SsaRaNormalizer.normTypeQueryRec() [aeneas/src/ir/SsaNormalizer.v3 @ 1074:66]
        in SsaRaNormalizer.normTypeQuery() [aeneas/src/ir/SsaNormalizer.v3 @ 1106:45]
        in SsaRaNormalizer.genApplyOp() [aeneas/src/ir/SsaNormalizer.v3 @ 171:62]
        in SsaRebuilder.genInstrs() [aeneas/src/ssa/SsaRebuilder.v3 @ 47:60]
        in SsaRebuilder.genBlock() [aeneas/src/ssa/SsaRebuilder.v3 @ 39:26]
        in SsaRebuilder.genMultiBlock() [aeneas/src/ssa/SsaRebuilder.v3 @ 30:53]
        in SsaRaNormalizer.genGraph() [aeneas/src/ir/SsaNormalizer.v3 @ 122:38]
        in SsaRaNormalizer.build() [aeneas/src/ir/SsaNormalizer.v3 @ 19:41]
        in ReachabilityNormalizer.normCode() [aeneas/src/ir/Normalization.v3 @ 812:57]
        in ReachabilityNormalizer.normalize() [aeneas/src/ir/Normalization.v3 @ 103:33]
        in ReachabilityAnalyzer.transform() [aeneas/src/ir/Reachability.v3 @ 64:37]
        in Compilation.reachability() [aeneas/src/main/Compiler.v3 @ 333:29]
        in Compilation.compile() [aeneas/src/main/Compiler.v3 @ 252:67]
[...]

PR #670, opened 2026-06-03
[unboxing] Fix Type Queries for Unboxed Tagless Variants

dnguyen-23:
Fix for https://github.com/titzer/virgil/issues/649
Unboxed Tagless Variants have a null VariantField. VariantNorm.tagType() doesn't case on the possibility of tag being null. VariantSolver slightly refactored to make tagging logic more clear.

### DIFF
--- a/aeneas/src/ir/SsaNormalizer.v3
+++ b/aeneas/src/ir/SsaNormalizer.v3
@@ -1110,11 +1110,7 @@ class SsaRaNormalizer extends SsaRebuilder {
 	}
 	// Returns whether `parent` VariantNorm has `child` as a VariantNorm in `parent.children`
 	def isVNParentChild(parent: VariantNorm, child: VariantNorm) -> bool {
-		for (l = parent.children; l != null; l = l.tail) {
-			def curChild: VariantNorm = l.head;
-			if (curChild == child) return true;
-		}
-		return false;
+ 		return Lists.find(parent.children, child) != -1;
 	}
 	def opAnd(left: SsaInstr, right: SsaInstr) -> SsaInstr {
 		return if(left == null, right, curBlock.opBoolAnd0(left, right));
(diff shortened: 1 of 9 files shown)


aeneas/src/ir/VariantNormalizer.v3
```diff
@@ -39,7 +39,7 @@ class VariantNorm extends TypeNorm {
 		return size == 1 && tag != null && fields.length == 0; 
 	}
 	def isEnumCase() -> bool { return size == 1 && tag != null && tagValue >= 0 && fields.length == 0; }
-	def tagType() -> IntType { return IntType.!(tag.tn.first()); }
+	def tagType() -> IntType { return if(tag == null, null, IntType.!(tag.tn.first())); }
```
titzer: This can be shorter as `if(tag != null, IntType.!(tag.tn.first()))`; i.e. the value for the condition being `false` can be omitted and it will be the default value for the expression type.

aeneas/src/ir/SsaNormalizer.v3
```diff
@@ -1105,6 +1108,14 @@ class SsaRaNormalizer extends SsaRebuilder {
 		}
 		return left;
 	}
+	// Returns whether `parent` VariantNorm has `child` as a VariantNorm in `parent.children`
+	def isVNParentChild(parent: VariantNorm, child: VariantNorm) -> bool {
+		for (l = parent.children; l != null; l = l.tail) {
```
titzer: I think it'd be nice to have a utility method `Lists.find() -> int` that returns the index of an element.

aeneas/src/ir/VariantNormalizer.v3
dnguyen-23: Resolved

aeneas/src/ir/SsaNormalizer.v3
dnguyen-23: Added the util function in the Lists component.

titzer (approved):
lgtm when CI is green again

PR #682, opened 2026-06-28
[enums] Added Enum set case to norm type method in Normalization

dnguyen-23:
Prior to this, encountering an enumset during normalization did not lead to errors. This PR is made for the sake of consistency.

### DIFF
--- a/aeneas/src/ir/Normalization.v3
+++ b/aeneas/src/ir/Normalization.v3
@@ -274,6 +274,9 @@ class ReachabilityNormalizer(config: NormalizerConfig, ra: ReachabilityAnalyzer)
 			ENUM => {
 				tn = TypeNorm.new(t, V3.getVariantTagType(t), null);
 			}
+			ENUM_SET => {
+				tn = TypeNorm.new(t, EnumSetType.!(t).repType, null);
+			}
 			REF => {
 				var sub = [
 					V3.arrayByteType,


titzer:
Are there any tests that you can think of that would expose a difference in behavior?

dnguyen-23:
I think no because: 

Case 1.) When the target has a MachProgram that uses MachLowering, the MachLowering.norm method already normalizes the EnumSetType. 

Case 2.) When the target is the JVM, JvmRep.v3 has a method called makeRep that performs the lowering for enum sets.

Case 3.) When the target is v3i, things become more sketchy. A problem will arise if the runtime value for an enum set instruction is not a Box<IntType>. I thought about all possible enum set initializations. 
Base case:
- `all`: bindVar -> `VarBinding.Const -> SsaConst` which has a `IntType` value corresponding to reptype.max
- default null initialization: MethodEnv.bindVar with `VarDecl.init == null` -> `SsaConst` that's a nullConst (`Box<IntType> == 0`)
- Union or Intersection of enums: `EnumSetType` defines a `genEnumToSet` method that represents the enum set as a series of bitwise operations on an enum input. However, we know that enums will be represented as a `Box<reptype>` 

Recursive case: (i.e union or intersection of enum sets, function calls)
- Assuming (by IH) two enum set expressions have a corresponding runtime Box<IntType> Value, then the union or intersection of them yields a new enum set expression that has a corresponding value.

titzer:
Ok, we can merge this without tests then.

PR #689, opened 2026-07-07
Guard against incorrect assumed operand sizes from `PtrAdd` and `PtrSub`

matthew-mojira:
Claude discovered this issue while it was writing some Virgil code for me. Testing first to see if CI will also fail based on my local tests..... From my testing it fails on -O0 and -O3 but -O0 is not in the CI


matthew-mojira:
The `PtrAdd` and `PtrSub` opcodes always use 8-byte assembly opcodes (`I_ADDQ` and `I_SUBQ`) to emit the instruction (in the call to `SsaX86_64Gen.emitSimpleBinop`). However, when a I32 is spilled to the stack it reserves an 8-byte stack slot without clearing the highest 4 bytes. When the instruction is emitted it as treated as a full 8-byte load from the stack, including the highest 4 bytes as garbage.

The test exposes this bug by performing pointer arithmetic on a Range access (`addr + offset`) with the offset as a value on the stack where the upper 4 bytes are previously overwritten with garbage (a lot of junk in the way as to prevent the optimizer from optimizing away the overwrite on the stack).

The fix proposed in the PR checks if the opcode expects 8-byte operands but the second vreg is an I32, in which case it forces that the operand is used in a register.

titzer:
Nice find! I think the real fix should maybe do this in assembling the instruction, because it's a problem with the addressing mode. Can you have claude write smaller tests that exhibit it? People have to also look at the asm :-)

matthew-mojira:
I beat Claude into making the tests smaller:
* [stale_spill_width00.v3](https://github.com/titzer/virgil/pull/689/changes#diff-96c6fe43fe045b007bb0208ebaa8a7290656557f57315622c48f064e77dda334) will fail on `-O0`
* [stale_spill_width01.v3](https://github.com/titzer/virgil/pull/689/changes#diff-5f0454cd095df2b52b23a43d4c4db7511ece2767afdd81c88ea97bdf73d4484f) will fail on `-O3`

titzer:
Superceded by #760

PR #693, opened 2026-07-14
Add `CiRuntime.readTsc`

matthew-mojira:
Compiles into the x86 `rdtsc` instruction and puts the full 64-bit result in `rax`


titzer:
This is probably done better with a machine-specific intrinsic. Closing for now.

PR #707, opened 2026-07-22
Restore pending shadow stack restores before shadow stack deallocation

matthew-mojira:
If a boxed value is saved to the shadow stack across two GC collections without being used in between (and there is straight line code between the two collections), then the new location of the value is not restored after the two GCs.

aa6e091a61f5a1afce4135f8363d53325d941894 removes an additional place where `ARCH_SETSHADOWSTACKPTR` is handled. I couldn't get Claude to come up with a repro around this point hence it's not included in the PR but might still be worth including.


test/rt/shadow_stack_livepoint0.v3
```diff
@@ -0,0 +1,19 @@
...
+// no intervening control flow is correctly reloaded from the shadow stack
+// after both calls, even if a collection relocates it during either call.
+class GBox {
+	def val: int;
+	new(val) { }
+}
+
+def add(t: GBox, a: GBox) -> GBox {
+	var r = GBox.new(t.val + a.val);
+	RiGc.forceGC();
```
titzer: Per discussion, I think you can use `CiTest.forceGc()` now and move this test into `test/gc`.

titzer (approved):

titzer:
Superceded by #762 . There were additional bugs in the MachLowering that tricked the liveness analysis for references.

PR #708, opened 2026-07-22
Add `finishLoopLiveness` to Wasm backend

matthew-mojira:
A boxed value that is live across a loop but not used within a loop is not saved to the shadow stack if there is a GC collection inside the loop. Mirrors the inclusion of `finishLoopLiveness` in [MachBackend.v3](https://github.com/matthew-mojira/virgil/blob/pr/fix_wasm_loop/aeneas/src/mach/MachBackend.v3)


titzer:
Nice catch. Can you also add some `test/execute` and/or `test/large` tests and make sure they end up in `test/core/*.gc`, which will run them with the GC fuzzing turned on? For that you need to allocate a few objects and cause collections that are meant to induce heap corruption.

titzer:
Gonna land this and have Claude write some more tests.

PR #717, opened 2026-07-24
Debug smoke tests

titzer:


titzer:
This was a terribly flawed experiment in how to structure bugfixes. It would have been a mess because the huge number of intermediate commits enabling tests that weren't really green would have made bisection a nightmare in the future. So we avoid the nightmare future and pursue a strategy where each commit is green.

PR #734, opened 2026-07-29
feat: Simplified VariantSolver logic

dnguyen-23:
Simplification and refactor of the VariantSolver logic to make it easier to debug. The packing logic flow has been isolated from the default unboxing logic.

### DIFF
--- a/aeneas/src/os/Darwin.v3
+++ b/aeneas/src/os/Darwin.v3
@@ -24,6 +24,7 @@ class DarwinTarget extends Target {
 			compiler.NormConfig.RangeStartType = it; // XXX: use unsigned for range start type?
 //			compiler.NormConfig.ArrayIndexType = it; // TODO: enable 64-bit array index start?
 		}
+		compiler.NormConfig.MaxScalarWidth = space.addressWidth;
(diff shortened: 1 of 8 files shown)


aeneas/src/os/Darwin.v3
```diff
@@ -24,6 +24,7 @@ class DarwinTarget extends Target {
 			compiler.NormConfig.RangeStartType = it; // XXX: use unsigned for range start type?
 //			compiler.NormConfig.ArrayIndexType = it; // TODO: enable 64-bit array index start?
 		}
+		compiler.NormConfig.MaxScalarWidth = space.addressWidth;
```
titzer: I think the various targets should use the `IntNormalizer`, rather than the address space size, since that configures how integer arithmetic is lowered.

titzer (approved):
Looks good with minor suggestion

aeneas/src/os/Darwin.v3
dnguyen-23: Resolved for Darwin, Linux, Wasm, and WasmGC. v3i and JVM use the default 64, if that's alright.

PR #737, opened 2026-07-29
Mostly this adds support for clwb and related instructions and more f…

eliotmoss:
…ences.

It also enhances the trace output a bit.


titzer (approved):
lgtm, rubber-stamped

PR #741, opened 2026-08-01
[packing] Reference handling in PackingSolver

dnguyen-23:
ub_ref00.v3 was the failing case. 

Problem: Packing references fails. Tagged reference operations are not supported operators in the compiler yet. 

Solution: Avoid packing fields into scalars for references. PackingSolver moves onto next scalar after assigning a ref to a scalar.

PR includes changes to PackingSolver to take in the NormalizerConfig so that the solver can case on the MaxScalarWidth in the future. This allows the solver to determine which reference to use for packing. Targets could modify the NormalizerConfig in their configureCompiler methods to flag whether tagged references are allowed for the target. As of right now, tagged references are supported in the runtime.

### DIFF
--- a/aeneas/src/ir/PackingSolver.v3
+++ b/aeneas/src/ir/PackingSolver.v3
@@ -21,12 +21,11 @@ def EMPTY_INTERVAL = Interval(0, 0);
 
 // Contains a number of routines for checking distinguishability of bit patterns and solving for
 // sets of patterns (by choosing assignments of bits).
-class PackingSolver(width: byte, refPatterns: RefPatterns) {
+class PackingSolver(width: byte, nc: NormalizerConfig, UNIFY: bool) {
 	// ----------------- Constants -----------------
 	var SCALAR_LIMIT = 2;
 	def DEBUG = CLOptions.PRINT_PACKING.get();
 	def REF_KEY = -1;
-	def SCALAR_WIDTH = if(refPatterns != null, refPatterns.ptrref.size, width);
 	// ----------------- Data structures -----------------
 	// Array of maps for each case/variant. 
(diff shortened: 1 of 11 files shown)


aeneas/src/ir/PackingSolver.v3
```diff
@@ -21,12 +21,11 @@ def EMPTY_INTERVAL = Interval(0, 0);
 
 // Contains a number of routines for checking distinguishability of bit patterns and solving for
 // sets of patterns (by choosing assignments of bits).
-class PackingSolver(width: byte, refPatterns: RefPatterns) {
+class PackingSolver(width: byte, nc: NormalizerConfig, UNIFY: bool) {
```
titzer: To reduce coupling, it's better if the PackingSolver takes what it needs from the normalizer config, rather than the normalizer config itself. The normalizer config is a big grab bag, and it's better if we see what exactly the solver depends directly.

test/variants/ub_ref00.v3
```diff
@@ -0,0 +1,25 @@
+//@execute 0=0; 1=1; 2=2; 3=3; 4=0; 5=1
```
titzer, line 1: It's always good to write at least a few tests that have the same bug that are all slightly different, to get more coverage. That way the bug doesn't get accidentally sidestepped. A little family of buggy programs :)

dnguyen-23: Added more tests

aeneas/src/ir/PackingSolver.v3
dnguyen-23: Passing in MaxScalarWidth instead of passing in the NormalizerConfig

titzer (approved):

PR #763, opened 2026-08-24
Annotations: syntax, parser, and verifier

eliotmoss:
First round of Claude-authored code for annotations.  Feedback welcome.  A later "true" PR will squash all of this ...

### DIFF
--- a/aeneas/src/main/Aeneas.v3
+++ b/aeneas/src/main/Aeneas.v3
@@ -61,6 +61,18 @@ component Aeneas {
 				break;
 			}
 		}
+		// Prepend the standard annotation declarations, when annotations are enabled.
+		// The compiler cannot find them itself: main() gets no argv[0] and there is no
+		// getenv, so the path comes from the command line, as -rt.files does.
+		if (CLOptions.ANNOTATIONS.get()) {
+			var afiles = CLOptions.ANNOTATION_FILES.get();
+			if (afiles.length > 0) {
+				var v = Vector<string>.new().grow(afiles.length + prog.files.length);
+				v.puta(prog.files);
+				v.puta(afiles);
+				prog.files = v.extract();
+			}
+		}
 		// Prepend -rt.files option to program files
 		var rtfiles = CLOptions.RT_FILES.get();
 		if (rtfiles.length > 0) {
--- a/aeneas/src/main/CLOptions.v3
+++ b/aeneas/src/main/CLOptions.v3
@@ -43,6 +43,8 @@ component CLOptions {
 		"Enable descriptors."), falseOption);
 	def ANNOTATIONS		= if(Debug.UNSTABLE, langOpt.newBoolOption("lang:annotations", false,
 		"Enable annotations."), falseOption);
+	def ANNOTATION_FILES	= langOpt.newOption("lang.annotation-files", Array<string>.new(0), "=<path*>", parsePathArray,
+		"Set a list of .v3 files declaring the standard annotations, included with every compiled program when -lang:annotations is on. Separate paths with commas or spaces.");
 	// Optimization and representation options.
 	def DESCEXT		= if(Debug.UNSTABLE, sharedOpt.newBoolOption("descext", false,
 		"Force described objects to use their descriptor as the object header.")
@@ -311,6 +313,23 @@ component CLOptions {
 				'n' => CLOptions.PRINT_INLINING.val = true;
 			}
 		}
+	}
+	// Like parseStringArray, but also accepts commas. A caller that cannot quote the value
+	// -- a shell variable that will be word-split, such as V3C_OPTS in the test harness --
+	// needs a separator that survives as a single argument.
+	def parsePathArray(str: string) -> Array<string> {
+		var v = Vector<string>.new();
+		var start = 0, pos = 0;
+		while (pos < str.length) {
+			var ch = str[pos];
+			if (ch == ' ' || ch == '\n' || ch == ',') {
+				if (pos > start) v.put(Arrays.range(str, start, pos));
+				start = pos + 1;
+			}
+			pos++;
+		}
+		if (pos > start) v.put(Arrays.range(str, start, pos));
+		return v.extract();
 	}
 	def parseStringArray(str: string) -> Array<string> {
 		var v = Vector<string>.new();
--- a/aeneas/src/main/CLOptions.v3
+++ b/aeneas/src/main/CLOptions.v3
@@ -41,6 +41,8 @@ component CLOptions {
 		"Enable simple function body syntax and return type inference.");
 	def DESCRIPTORS	= if(Debug.UNSTABLE, langOpt.newBoolOption("lang:descriptors", false,
 		"Enable descriptors."), falseOption);
+	def ANNOTATIONS		= if(Debug.UNSTABLE, langOpt.newBoolOption("lang:annotations", false,
+		"Enable annotations."), falseOption);
 	// Optimization and representation options.
 	def DESCEXT		= if(Debug.UNSTABLE, sharedOpt.newBoolOption("descext", false,
 		"Force described objects to use their descriptor as the object header.")
--- a/aeneas/src/main/Compiler.v3
+++ b/aeneas/src/main/Compiler.v3
@@ -278,6 +278,8 @@ class Compilation(compiler: Compiler, prog: Program) {
 			}
 			vec.put(Parser.parseFile(fileName, input, prog.ERROR, prog.typeCache));
 		}
+		// Desugar annotation declarations into open variant types before verification.
+		if (CLOptions.ANNOTATIONS.get()) AnnotationDesugaring.desugar(vec);
 		prog.vst.files = vec.extract();
 		if (CLOptions.REDEF_FIELD.val != null) {
 			var input = StringBuilder.new().puts("-").puts(CLOptions.REDEF_FIELD.name).puts("=");
--- a/aeneas/src/test/Regression.v3
+++ b/aeneas/src/test/Regression.v3
@@ -113,12 +113,29 @@ class Regression {
 		}
 		return null;
 	}
+	// Append the standard annotation declarations, as Aeneas.makeProgram does for a normal
+	// compile. Test programs are assembled here rather than there, so the option would
+	// otherwise have no effect under -test. A null input means "load from disk".
+	def addAnnotationFiles(prog: Program) -> Program {
+		if (!CLOptions.ANNOTATIONS.get()) return prog;
+		var afiles = CLOptions.ANNOTATION_FILES.get();
+		if (afiles.length == 0) return prog;
+		var f = Vector<string>.new().puta(prog.files);
+		var i = Vector<Array<byte>>.new().puta(prog.inputs);
+		for (j < afiles.length) {
+			f.put(afiles[j]);
+			i.put(null);
+		}
+		prog.files = f.extract();
+		prog.inputs = i.extract();
+		return prog;
+	}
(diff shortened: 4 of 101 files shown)


eliotmoss:
I should add that doc has several markdown documents describing things in detail, along the stages by which it was implemented.

aeneas/src/main/Aeneas.v3
```diff
@@ -61,6 +61,18 @@ component Aeneas {
 				break;
 			}
 		}
+		// Prepend the standard annotation declarations, when annotations are enabled.
```
titzer, line 64: Probably time to factor out a subroutine for adding files to the program that's used for RT and annotations.

aeneas/src/main/CLOptions.v3
```diff
@@ -310,6 +314,23 @@ component CLOptions {
 			}
 		}
 	}
+	// Like parseStringArray, but also accepts commas. A caller that cannot quote the value
```
titzer, line 317: I think it's better to just keep the same convention as RT files and not use commas for now.

aeneas/src/main/Compiler.v3
```diff
@@ -278,6 +278,8 @@ class Compilation(compiler: Compiler, prog: Program) {
 			}
 			vec.put(Parser.parseFile(fileName, input, prog.ERROR, prog.typeCache));
 		}
+		// Desugar annotation declarations into open variant types before verification.
+		if (CLOptions.ANNOTATIONS.get()) AnnotationDesugaring.desugar(vec);
```
titzer, line 282: This makes a pass over ever file and declaration; I think it's probably better to make sure the program has a complete index either by the end of parsing or at the start of verification, where other declarations are collected.

aeneas/src/test/Regression.v3
```diff
@@ -113,12 +113,29 @@ class Regression {
 		}
 		return null;
 	}
+	// Append the standard annotation declarations, as Aeneas.makeProgram does for a normal
+	// compile. Test programs are assembled here rather than there, so the option would
+	// otherwise have no effect under -test. A null input means "load from disk".
+	def addAnnotationFiles(prog: Program) -> Program {
```
titzer, line 119: This looks like a duplication that could be factored out.

aeneas/src/vst/AnnotationDesugaring.v3
```diff
@@ -0,0 +1,153 @@
+// Copyright 2026 Virgil authors. All rights reserved.
+// See LICENSE for details of Apache 2.0 license.
+
+// Desugars annotation type declarations into open variant types.
+//
+// Each "@type" declaration becomes a synthesized variant extending a synthesized open
```
titzer, line 6: These are very characteristic long Claude comments. Style wise I prefer much more targeted comments that are more likely to be read and don't take up a lot of real estate.

aeneas/src/vst/AnnotationEnv.v3
```diff
@@ -0,0 +1,163 @@
+// Copyright 2026 Virgil authors. All rights reserved.
+// See LICENSE for details of Apache 2.0 license.
+
+// What a use of "@name" resolves to.
+type AnnotationBinding {
```
titzer, line 5: This seems like a way to adapt two data types that have basically the same members to the same interface. It'd be better to factor out the common data to be shared by both so that no adapter is necessary.

aeneas/src/vst/AnnotationEnv.v3
```diff
@@ -0,0 +1,163 @@
...
+		// would be misleading rather than merely redundant.
+		if (Strings.equal(name, "file") || Strings.equal(name, "apply")) {
+			err(b.file(), token).set("AnnotationDeclError",
+				Strings.format1("\"@%s\" is reserved and cannot be declared", name));
+			return;
+		}
+		checkOverlap(name, b, token);
+		env.add(name, b);
+	}
+	// Two declarations of one name are legal only for disjoint target sets. Deciding that
```
titzer, line 109: AI likes to make these kind of high-level semantics documentation on private methods, which is an anti-pattern that makes private things really important and also hides it. Generally I don't like documenting language semantics in the compiler. It gets out of date and is usually a distraction. No one should be looking at the compiler to find out language semantics. Unless it's something very subtle explaining some complicated corner case, and even then it's better to keep it short.

aeneas/src/vst/AnnotationEval.v3
```diff
@@ -0,0 +1,223 @@
...
+			var e = a.boundArgs[i];
+			if (e == null) continue;	// absent; the field takes its type's default
+			var r = eval(e);
+			if (r.0) vals[i] = r.1;
+			else err(a, e);
+		}
+		a.values = vals;
+	}
+	// Returns (ok, value).
+	def eval(e: Expr) -> (bool, Val) {
```
titzer, line 68: Hmm, this introduces an evaluator for a subset of the AST in a way that makes me unsure if it will grow in complexity.

aeneas/src/vst/AnnotationEval.v3
```diff
@@ -0,0 +1,223 @@
...
+		}
+		args.reset(op, vals);
+		var r = Eval.doOp(op, args);
+		if (args.failed || r == null) return (false, null);
+		match (r) {
+			v: Val => return (true, v);
+			_ => return (false, null);	// an exception is not a constant
+		}
+	}
+	private def evalSugar(vst: VstOperator, vals: Array<Val>, at: Expr) -> (bool, Val) {
```
titzer, line 166: Oh, an evaluator for sugar operators. That definitely feels like it's going to be brittle.

aeneas/src/vst/AnnotationResolver.v3
```diff
@@ -0,0 +1,225 @@
...
+// Where an annotation is attached. Mirrors the TargetKind enum in
+// lib/annotations/TargetKind.v3, which is what a user writes in "@target(...)"; this is
+// the compiler's own copy, because a site must be classified whether or not the standard
+// annotations have been loaded.
+enum AnnotationTarget {
+	Program, File, AnnotationType, AnnotationField,
+	Class, Component, Variant, Enum, VariantCase, EnumCase,
+	Field, Method, Constructor, Parameter,
+	TypeUse, Block,
+	// Not accepted by the parser; present so the two enums correspond.
```
titzer, line 13: I think it's better to leave out things that aren't supported yet; they just clutter the code with stuff that's not implemented yet.

aeneas/src/vst/AnnotationResolver.v3
```diff
@@ -0,0 +1,225 @@
...
+		if (list == null) return;
+		for (l = list.list; l != null; l = l.tail) resolve(file, l.head, target);
+		checkRepeats(file, list);
+	}
+	// "Attached twice at the same position" is exactly one list, and whether a declaration
+	// is @repeatable is syntactic, so this needs no evaluation.
+	private def checkRepeats(file: VstFile, list: VstList<VstAnnotation>) {
+		for (l = list.list; l != null; l = l.tail) {
+			for (m = l.tail; m != null; m = m.tail) {
+				if (!Strings.equal(l.head.name.image, m.head.name.image)) continue;
```
titzer, line 113: AI loves string comparison. In the Virgil compiler, string comparisons are mostly relegated to verification, where binding happens. This should check for repeated bindings, not names.

aeneas/src/vst/AnnotationResolver.v3
```diff
@@ -0,0 +1,225 @@
...
+	private def isRepeatable(b: AnnotationBinding) -> bool {
+		var anns: VstList<VstAnnotation>;
+		match (b) {
+			Decl(d) => anns = d.annotations;
+			Alias(a) => anns = a.annotations;
+			_ => return true; // unresolved; do not pile on
+		}
+		if (anns == null) return false;
+		for (l = anns.list; l != null; l = l.tail) {
+			if (Strings.equal(l.head.name.image, "repeatable")) return true;
```
titzer, line 130: More string comparisons for semantically important things.

aeneas/src/vst/Parser.v3
```diff
@@ -125,10 +126,111 @@ component Parser {
...
+			decl.file = file;
+			file.annotationDecls.put(decl);
+			return true;
+		}
+		if (optKeyword(p, "def") != null) {
+			var id = parseIdentVoid(p).name;
+			p.req1('=');
+			var start = p.point();
+			var list: List<VstAnnotation>;
+			while (true) {
```
titzer, line 175: There is a parseList for this type of situation.

aeneas/src/vst/Parser.v3
```diff
@@ -306,9 +417,81 @@ component Parser {
...
+			var a = parseAnnotation(p);
+			if (a == null) break;
+			list = List.new(a, list);
+		}
+		if (list == null) return null;
+		return VstList.new(p.end(start), Lists.reverse(list));
+	}
+	// Is the identifier immediately after the "@" a keyword? Scans the raw input without
+	// moving the cursor, so that "@type" and "@def" are left for the caller.
+	def isKeywordAfterAt(p: ParserState) -> bool {
```
titzer, line 441: This is the kind of scanner thing that seems very brittle. It's poorly factored and won't handle whitespace right.

aeneas/src/vst/Parser.v3
```diff
@@ -306,9 +417,81 @@ component Parser {
...
+			p.error("annotation name expected");
+			return null;
+		}
+		// Scan the name directly rather than via parseIdentCommon, which would take a
+		// following '<' as type arguments; annotation names are not parameterized.
+		var q = p.star(0, Char.isIdentMiddle);
+		// The argument list must be glued to the name: "@ann(x)" is arguments, while
+		// "@ann (x)" applies @ann to the parenthesized type or expression. This must be
+		// read from the raw input BEFORE extractIdent, which advances past whitespace.
+		var glued = q < p.input.length && p.input[q] == '(';
```
titzer, line 466: Also very low-level character checking; it doesn't handle whitespace correctly.

aeneas/src/vst/Parser.v3
```diff
@@ -306,9 +417,81 @@ component Parser {
...
+		}
+		return VstAnnArg.new(null, e);
+	}
+	// An annotation used as a value, plus any suffixes: "@outer.@atomicity.value".
+	def parseAnnValue(p: ParserState) -> Expr {
+		var a = parseAnnotation(p);
+		if (a == null) return null;
+		return addBinOpSuffixes(p, addTermSuffixes(p, AnnotationExpr.new(a)));
+	}
+	def isBareVarExpr(e: Expr) -> bool {
```
titzer, line 490: This is making decisions based on the shape of an already parsed AST node; it should instead do this the way the rest of the parser does; with control flow or boolean flags.

aeneas/src/vst/Parser.v3
```diff
@@ -306,9 +417,81 @@ component Parser {
...
+		// read from the raw input BEFORE extractIdent, which advances past whitespace.
+		var glued = q < p.input.length && p.input[q] == '(';
+		var id = extractIdent<void>(p, q);
+		var args = if(glued, parseList(0, p, '(', COMMA, ')', parseAnnArg));
+		return VstAnnotation.new(at, id.name, args);
+	}
+	// Parse one annotation argument, either "name = value" or a bare value.
+	def parseAnnArg(p: ParserState) -> VstAnnArg {
+		var e = if(p.curByte == '@', parseAnnValue(p), parseExpr(p));
+		if (AssignExpr.?(e)) {
```
titzer, line 474: This is matching on an already-parsed VST node.

PR #777, opened 2026-09-15
Add `#inline-into` compiler directive

matthew-mojira:
The `#inline-into` directive is like the opposite of the `#inline` directive. When a function is declared as `#inline-into` the compiler forces any functions/methods that it calls to be inlined into that function. Would be useful for fast handler functions to force the compiler to reduce the number of calls in these hot functions.


titzer (approved):
The flag and the behavior makes sense, but I'm not sure that inlining actually got activated by this. There isn't currently a way to assert optimizations, but did you inspect code manually?

Also, the compiler doesn't currently support polymorphic inlining, but that'd be good to have as well, since we could replace a `Call*Virtual` with a cascade of type cases to direct calls, and inline each of them.

